PR #47, opened 2023-10-27
added multiple positions per same tick

Sniezka1927:


src/lib.rs
```diff
@@ -179,19 +179,20 @@ pub mod contract {
 
             let pool = self.pools.get_pool(pool_key)?;
 
-            let tick_option = self.ticks.get_tick(pool_key, index);
-            if tick_option.is_some() {
-                return Err(ContractErrors::TickAlreadyExist);
+            let tick = self.ticks.get_tick(pool_key, index);
```
wojciech-cichocki, line 182: I don't think this is the best way to solve this problem. 

What is the actual cause of the problem?
```rust
let mut upper_tick = self
    .ticks
    .get_tick(pool_key, upper_tick)
    .unwrap_or(self.create_tick(pool_key, upper_tick)?);
```
The reason for this is the mechanism of `unwrap_or` [link](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or). So, this code, `self.create_tick(pool_key, upper_tick)?`, is eagerly evaluated, which means that the tick will be created first, and then it will be checked whether the tick is created. The solution is to use `unwrap_or_else` for lazy evaluation. This means that the tick creation code will only trigger after checking if the tick is not already created.

You need to update all non-static examples of invoking `unwrap_or` to `unwrap_or_else`:
```rust
let mut lower_tick = self
  .ticks
  .get_tick(pool_key, lower_tick)
  .unwrap_or_else(|| Self::create_tick(self, pool_key, lower_tick).unwrap());
```

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #49, opened 2023-10-30
refactor pools and create pool

zielvna:


src/lib.rs
```diff
@@ -678,18 +678,23 @@ pub mod contract {
             fee_tier: FeeTier,
             init_tick: i32,
         ) -> Result<(), ContractErrors> {
-            let current_timestamp = self.env().block_timestamp();
-
-            let key = FeeTierKey(fee_tier.fee, fee_tier.tick_spacing);
+            if token_0 == token_1 {
```
wojciech-cichocki, line 681: It will be better to move this validation to `PoolKey::new`

PR #53, opened 2023-10-31
Add claim test

zielvna:


src/contracts/collections/positions.rs
```diff
@@ -22,6 +22,13 @@ impl Positions {
             .insert(account_id, &(positions_length, positions));
     }
 
+    pub fn update(&mut self, account_id: AccountId, index: u32, position: &Position) {
```
wojciech-cichocki, line 25: throw Error if position not exists

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #56, opened 2023-11-02
Add max tick cross test

zielvna:


src/lib.rs
```diff
@@ -931,6 +931,94 @@ pub mod contract {
...
+                dex,
+                token_x,
+                token_y,
+                pool_key.fee_tier
+            )
+            .unwrap();
+
+            let swap_amount = TokenAmount::new(amount);
+            let slippage = SqrtPrice::new(MIN_SQRT_PRICE);
+            swap!(
```
wojciech-cichocki, line 993: It will be additional check if you check quote entrypoint also here and validate result like: crossed tick etc

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #61, opened 2023-11-03
Add limits test

zielvna:


src/lib.rs
```diff
@@ -910,29 +910,314 @@ pub mod contract {
...
 
         type E2EResult<T> = Result<T, Box<dyn std::error::Error>>;
 
+        #[ink_e2e::test]
+        #[should_panic]
+        async fn limits_swap_at_upper_limit(mut client: ink_e2e::Client<C, E>) -> () {
+            let (dex, token_x, token_y) =
+                init_dex_and_tokens_max_mint_amount!(client, ContractRef, TokenRef);
+
+            let mint_amount = 2u128.pow(63) - 1;
```
wojciech-cichocki, line 947: This test should be aligned. Max TokenAmount is `2^128`, not` 2^64`

wojciech-cichocki (changes requested):

src/contracts/logic/math.rs
```diff
@@ -230,6 +230,10 @@ pub fn calculate_y(
     })
 }
 
+pub fn get_max_tick(tick_spacing: u16) -> i32 {
```
wojciech-cichocki, line 233: this function is in math

wojciech-cichocki (changes requested):
Add also full range with max liquidity test case

wojciech-cichocki (changes requested):

src/lib.rs
```diff
@@ -910,44 +910,354 @@ pub mod contract {
...
+                TokenAmount(pos_amount),
+                lower_tick,
+                upper_tick,
+                pool.sqrt_price,
+                false,
+            )
+            .unwrap()
+            .l;
+
+            let liquidity_delta = if liquidity_by_y < liquidity_by_x {
```
wojciech-cichocki, line 1104: You also need to update this case:
1. Calculate liquidity for one of the tokens.
2. Calculate the amount of the other token based on the liquidity calculated in point 1.
3. Add similar validation to contract balances - they should be exactly the same as the assumed amounts in points 1 and 2.

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #62, opened 2023-11-06
Cross both side spec

Sniezka1927:


src/lib.rs
```diff
@@ -1157,7 +1156,7 @@ pub mod contract {
             let swap_amount = TokenAmount::new(amount);
             approve!(client, TokenRef, token_x, dex, amount, alice);
 
-            let target_sqrt_price = SqrtPrice::new(990157831158286630888963);
+            let target_sqrt_price = SqrtPrice::new(MAX_SQRT_PRICE);
```
wojciech-cichocki, line 1159: It's `slippage_test`, so I am not sure whether `MAX_SQRT_PRICE` is the proper target. You probably should use something closer to the resulting sqrt price

src/lib.rs
```diff
@@ -1180,7 +1179,7 @@ pub mod contract {
                 target_sqrt_price,
                 alice
             );
-            let expected_sqrt_price = SqrtPrice::new(990157831158286630888964); // -197 tick
+            let expected_sqrt_price = SqrtPrice::new(1009940000000000000000000);
```
wojciech-cichocki, line 1182: similar case here

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #67, opened 2023-11-20
Added events structs

Sniezka1927:


wojciech-cichocki (changes requested):
building failed

src/contracts/storage/pool.rs
```diff
@@ -186,7 +186,8 @@ impl Pool {
         total_amount_in: &mut TokenAmount,
         protocol_fee: Percentage,
         fee_tier: FeeTier,
-    ) {
+    ) -> (bool, Option<i32>) {
+        let mut has_crossed = (false, None);
```
wojciech-cichocki, line 190: if case tick was crossed we know which one. `tick_index` is stored in `limiting_tick`

src/lib.rs
```diff
@@ -56,6 +57,52 @@ pub mod contract {
     use ink::prelude::vec::Vec;
     use token::PSP22;
 
+    #[ink(event)]
+    pub struct CreatePositionEvent {
```
wojciech-cichocki, line 61: you can move these structs to separate file such as `storage/events.rs`

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #68, opened 2023-11-20
add swap_route message

zielvna:


src/lib.rs
```diff
@@ -69,6 +69,19 @@ pub mod contract {
...
+    #[derive(scale::Decode, Default, scale::Encode, Clone, Debug)]
+    #[cfg_attr(
+        feature = "std",
+        derive(scale_info::TypeInfo, ink::storage::traits::StorageLayout,)
+    )]
+    pub struct SwapParams {
+        pool_key: PoolKey,
+        x_to_y: bool,
+        amount: TokenAmount,
+        by_amount_in: bool,
```
wojciech-cichocki, line 81: enable `by_amount_in` only

src/lib.rs
```diff
@@ -69,6 +69,19 @@ pub mod contract {
...
 
+    #[derive(scale::Decode, Default, scale::Encode, Clone, Debug)]
+    #[cfg_attr(
+        feature = "std",
+        derive(scale_info::TypeInfo, ink::storage::traits::StorageLayout,)
+    )]
+    pub struct SwapParams {
+        pool_key: PoolKey,
+        x_to_y: bool,
+        amount: TokenAmount,
```
wojciech-cichocki, line 80: Caller specifies `TokenAmount` only once. Then, the output amount of one hop is used as the input amount for another

src/lib.rs
```diff
@@ -69,6 +69,19 @@ pub mod contract {
...
+        by_amount_in: bool,
+        sqrt_price_limit: SqrtPrice,
```
wojciech-cichocki, line 82: There should be one limit. Set all limits to the maximum, and finally, check slippage based on the number of expected last hop tokens

wojciech-cichocki (changes requested):
There should also be e2e test. Swap involving 2 hops is sufficient

wojciech-cichocki (changes requested):
Resolve conflicts are required. Swap should emit the event, swap route not

wojciech-cichocki (approved):

PR #69, opened 2023-11-24
change positions collection structure and update methods

zielvna:


src/contracts/collections/positions.rs
```diff
@@ -46,18 +47,28 @@ impl Positions {
...
-            self.positions
-                .insert(account_id, &(positions_length, positions));
-            Ok(position)
+        let positions_length = self.get_length(account_id);
+
+        if index >= positions_length {
+            return Err(ContractErrors::PositionNotFound);
+        }
+
+        let position = self.positions.get((account_id, index)).unwrap_or_default();
```
wojciech-cichocki, line 56: It would be better to use `unwrap`. Although it has the same effect now, unwrap offers higher fault tolerance

wojciech-cichocki: or even return `PositionNotFound`

wojciech-cichocki (changes requested):
this task definitely requires e2e acceptance

src/contracts/collections/positions.rs
```diff
@@ -46,54 +45,200 @@ impl Positions {
...
+        let position = Position::default();
+        let new_position = Position {
+            lower_tick_index: -1,
+            upper_tick_index: 1,
+            ..Position::default()
+        };
+
+        positions.add(account_id, position);
+
+        let result = positions.update(account_id, 0, &new_position);
```
wojciech-cichocki, line 135: you can use `unwrap` in tests. Asserts in such cases only add irrelevant lines to the code.
`assert_eq!(result, Ok(()));`

src/contracts/collections/positions.rs
```diff
@@ -46,54 +45,200 @@ impl Positions {
...
+    #[ink::test]
+    fn test_add() {
+        let positions = &mut Positions::default();
+        let account_id = AccountId::from([0x01; 32]);
+        let position = Position::default();
+
+        positions.add(account_id, position);
+
+        assert_eq!(positions.get(account_id, 0), Some(position));
+        assert_eq!(positions.get(account_id, 1), None);
```
wojciech-cichocki, line 119: Initialization is the first case, and adding the next element is the next case to test. So, add here another `positions.add`
(I know that next tests may have more than one addition, but it is still worth testing it in the tests dedicated to adding)

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):
Overall, IMO `get_length` should be validated also in other test. Position length is inseparably linked with operations

PR #72, opened 2023-11-27
added creating 85 positions

Sniezka1927:


wojciech-cichocki (changes requested):
Fix SecondsPerLiquidty in this PR

wojciech-cichocki (changes requested):
You also need to update all spot when timestamp is stored and convert these values to seconds. For instance when tick is initialized

src/math/types/seconds_per_liquidity.rs
```diff
@@ -140,7 +140,7 @@ pub mod tests {
                 .unwrap();
             assert_eq!(
                 seconds_per_liquidity.get(),
-                315360000000000000000000000000000000000
+                315360000000000000000000000000000000
```
wojciech-cichocki, line 143: this doesn't look like max value: `log2(315360000000000000000000000000000000) < 2^118`

src/lib.rs
```diff
@@ -1213,6 +1213,63 @@ pub mod contract {
 
         type E2EResult<T> = Result<T, Box<dyn std::error::Error>>;
 
+        #[ink_e2e::test]
+        async fn timestamp_err(mut client: ink_e2e::Client<C, E>) -> E2EResult<()> {
```
wojciech-cichocki, line 1217: change name of this test to `test_calculating_seconds_per_liquidity`

wojciech-cichocki (changes requested):

PR #85, opened 2023-11-28
Refactor e2e

Sniezka1927:


src/lib.rs
```diff
@@ -1213,6 +1213,153 @@ pub mod contract {
 
         type E2EResult<T> = Result<T, Box<dyn std::error::Error>>;
 
+        #[ink_e2e::test]
+        async fn reversed(mut client: ink_e2e::Client<C, E>) -> E2EResult<()> {
```
wojciech-cichocki, line 1217: name it : `test_swap y_to_x` and similar case but in different direction `test_swap_x_to_y`

src/lib.rs
```diff
@@ -1405,10 +1547,18 @@ pub mod contract {
                 slippage_limit_upper,
                 alice
             );
+
+            let contract_amount_x = dex_balance!(TokenRef, client, token_x, dex);
+            let contract_amount_y = dex_balance!(TokenRef, client, token_y, dex);
+
+            let expected_x = 0;
+            let expected_y = 42534896005851865508212194815854;
```
wojciech-cichocki, line 1555: Add comment here with `< 2^106`

src/lib.rs
```diff
@@ -1764,7 +1905,7 @@ pub mod contract {
             let crosses_after_quote =
                 ((pool_after_quote.current_tick_index - pool_before.current_tick_index) / 10).abs();
             assert_eq!(crosses_after_quote, 0);
-            assert_eq!(quote_result.3.len() - 1, 19);
+            assert_eq!(quote_result.3.len() - 1, 146);
```
wojciech-cichocki, line 1908: quote_result should be definitely struct instead of tuple. It's hard to read what means `.3` , `.2`

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #86, opened 2023-11-28
Added entrypoint descriptions

Sniezka1927:


wojciech-cichocki (changes requested):
Describe parameters. BTW entyrpoints explanation is docs can definitely be improved. Check example: https://github.com/Cardinal-Cryptography/wAZERO/blob/main/traits.rs

src/contracts/entrypoints.rs
```diff
@@ -12,22 +12,63 @@ use ink::primitives::AccountId;
...
+    ///
+    /// # Parameters
+    /// - `pool_key`: A unique key that identifies the specified pool.
+    ///
+    /// # Errors
+    /// - Reverts the call when the caller is an unauthorized receiver.
     #[ink(message)]
     fn withdraw_protocol_fee(&mut self, pool_key: PoolKey) -> Result<(), InvariantError>;
 
+    /// Allows an authorized user to adjust the protocol fee.
```
zielvna, line 29: the only user that is allowed to change protocol fee is an admin so to be more specific "authorized user" should be changed to "admin"

src/contracts/entrypoints.rs
```diff
@@ -12,22 +12,63 @@ use ink::primitives::AccountId;
...
+    ///
+    /// # Parameters
+    /// - `protocol_fee`: The expected fee represented as a percentage.
+    ///
+    /// # Errors
+    /// - Reverts the call when the caller is an unauthorized user.
     #[ink(message)]
     fn change_protocol_fee(&mut self, protocol_fee: Percentage) -> Result<(), InvariantError>;
 
+    /// Transfers fee receiver authorization to another user.
```
zielvna, line 39: people reading that might think that fee receiver can also transfer their role to another user, you should be more specific

src/contracts/entrypoints.rs
```diff
@@ -116,6 +271,15 @@ pub trait Invariant {
...
 
+    /// Retrieves information about a pool created on a specified token pair with an associated fee tier.
+    ///
+    /// # Parameters
+    /// - `token_0`: The address of the first token.
+    /// - `token_1`: The address of the second token.
+    /// - `fee_tier`: A struct identifying the pool fee and tick spacing.
+    ///
+    /// # Errors
+    /// - Fails if the Pool key cannot be created.
```
zielvna, line 282: also fails if the pool with created pool key does not exist

src/contracts/entrypoints.rs
```diff
@@ -91,22 +199,69 @@ pub trait Invariant {
...
+    ///
+    /// # Parameters
+    /// - `token_0`: The address of the first token.
+    /// - `token_1`: The address of the second token.
+    /// - `fee_tier`: A struct identifying the pool fee and tick spacing.
+    /// - `init_tick`: The initial tick at which the pool will be created.
+    ///
+    /// # Errors
+    /// - Fails if the specified fee tier cannot be found.
+    /// - Fails if the user attempts to create a pool for the same tokens.
```
zielvna, line 264: should fail if pool with the same tokens and fee tier already exist (you should also refactor create pool message)

src/contracts/entrypoints.rs
```diff
@@ -91,22 +199,69 @@ pub trait Invariant {
...
+    ///
+    /// # Parameters
+    /// - `index`: The index of the user position to be removed.
+    ///
+    /// # Events
+    /// - Emits a `Remove Position` event upon success.
+    ///
+    /// # Errors
+    /// - Fails if unable to deinitialize ticks.
+    /// - Fails if the DEX has insufficient balance to perform the transfer.
```
zielvna, line 223: also fails if position can not be found

zielvna (approved):

wojciech-cichocki (changes requested):
"External contracts" can be added as another section (Parameters, Events, Errors). There should only be PSP22 here. However it will give a lot of information

wojciech-cichocki (approved):

PR #92, opened 2023-12-01
Refactor

Sniezka1927:


src/lib.rs
```diff
@@ -256,35 +256,30 @@ pub mod contract {
...
+                            &mut remaining_amount,
+                            by_amount_in,
+                            x_to_y,
+                            current_timestamp,
+                            &mut total_amount_in,
+                            self.state.protocol_fee,
+                            pool_key.fee_tier,
+                        );
+                        if has_crossed {
+                            self.emit_cross_tick_event(caller, pool_key, tick_index);
```
wojciech-cichocki, line 276: emitting events should be outside `calculate_swap` function. Now even quote message emitting cross events.

src/contracts/storage/pool.rs
```diff
@@ -174,11 +174,19 @@ impl Pool {
         ))
     }
 
+    pub fn update_tick_index(&mut self, result: SwapResult, fee_tier: FeeTier) {
```
wojciech-cichocki, line 177: What is the reason for defining this function? If this function is genuinely necessary, the naming should definitely be different. It updates the tick, but only in one of the possible cases

src/contracts/storage/pool.rs
```diff
@@ -200,20 +208,16 @@ impl Pool {
...
-                            .unwrap();
-                        *total_amount_in += *remaining_amount
-                    }
-                    *remaining_amount = TokenAmount(0);
+            if !x_to_y || is_enough_amount_to_cross {
+                let _ = tick.cross(self, current_timestamp);
+                has_crossed = true;
+            } else if !remaining_amount.is_zero() {
+                if by_amount_in {
+                    self.add_fee(*remaining_amount, x_to_y, protocol_fee)
```
wojciech-cichocki, line 216: use use `unwrap!` instead of unwrap

src/lib.rs
```diff
@@ -256,35 +256,30 @@ pub mod contract {
...
+                        let has_crossed = pool.cross_tick(
+                            result,
+                            swap_limit,
+                            tick_index,
+                            &mut tick,
+                            &mut remaining_amount,
+                            by_amount_in,
+                            x_to_y,
+                            current_timestamp,
+                            &mut total_amount_in,
```
wojciech-cichocki, line 271: return new `total_amount_in` instead of mutating it inside function. This is a bad practice

src/lib.rs
```diff
@@ -256,35 +256,30 @@ pub mod contract {
...
-                // TODO: refactor
-                let mut tick = Tick::default();
-
-                let update_limiting_tick = limiting_tick.map(|(index, bool)| {
-                    if bool {
-                        tick = self.ticks.get(pool_key, index).unwrap();
-                        (index, Some(&mut tick))
-                    } else {
-                        (index, None)
+                if let Some((tick_index, is_initialized)) = limiting_tick {
```
wojciech-cichocki, line 259: It's hard to determine if it's correct based on github changes, but are you sure if this logic is translated 1:1?

wojciech-cichocki (changes requested):

src/contracts/storage/pool.rs
```diff
@@ -129,61 +129,64 @@ impl Pool {
         }
     }
 
+    pub fn update_current_tick_index_when_deinitialized(
```
wojciech-cichocki, line 132: IMO this wrapper function should be removed
​

src/lib.rs
```diff
@@ -113,7 +113,7 @@ pub mod contract {
         pub target_sqrt_price: SqrtPrice,
         pub fee: TokenAmount,
         pub pool: Pool,
-        pub ticks: Vec<Tick>,
+        pub ticks: Vec<(Tick, bool)>,
```
wojciech-cichocki, line 116: compute swap step should return only crossed ticks. no crossing ticks don't matter

src/lib.rs
```diff
@@ -124,7 +124,7 @@ pub mod contract {
         pub amount_in: TokenAmount,
         pub amount_out: TokenAmount,
         pub target_sqrt_price: SqrtPrice,
-        pub ticks: Vec<Tick>,
+        pub ticks: Vec<(Tick, bool)>,
```
wojciech-cichocki, line 127: similar here. Only crossed (initalized) tick

wojciech-cichocki (changes requested):

src/lib.rs
```diff
@@ -551,8 +543,11 @@ pub mod contract {
             let calculate_swap_result =
                 self.calculate_swap(pool_key, x_to_y, amount, by_amount_in, sqrt_price_limit)?;
 
-            for tick in calculate_swap_result.ticks.iter() {
+            for (tick, has_crossed) in calculate_swap_result.ticks.iter() {
                 self.ticks.update(pool_key, tick.index, tick)?;
+                if *has_crossed {
+                    self.emit_cross_tick_event(caller, pool_key, tick.index);
```
wojciech-cichocki, line 549: also emit event once with all crossed ticks

wojciech-cichocki (changes requested):

src/lib.rs
```diff
@@ -551,10 +550,15 @@ pub mod contract {
...
                 self.calculate_swap(pool_key, x_to_y, amount, by_amount_in, sqrt_price_limit)?;
 
+            let mut crossed_tick_indexes: Vec<i32> = vec![];
+
             for tick in calculate_swap_result.ticks.iter() {
                 self.ticks.update(pool_key, tick.index, tick)?;
+                crossed_tick_indexes.push(tick.index);
             }
 
+            self.emit_cross_tick_event(caller, pool_key, crossed_tick_indexes);
```
wojciech-cichocki, line 560: do not emit event in case there was no cross. statistically >95% of swaps have no cross

PR #95, opened 2023-12-01
Group e2e tests

zielvna:


wojciech-cichocki (changes requested):
1. You can remove the separation of entrypoints and integration in the tests. This separation is incorrect anyway.
2. Of course, not all tests have been moved. So the PR is not ready.

PR #96, opened 2023-12-04
Removed clippy warnings

Sniezka1927:


src/lib.rs
```diff
@@ -300,6 +299,7 @@ pub mod contract {
             })
         }
 
+        #[allow(dead_code)]
```
wojciech-cichocki, line 302: add here implementation of remove tick

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #108, opened 2023-12-11
Extended claim fee tests

Sniezka1927:


src/e2e/claim.rs
```diff
@@ -55,4 +55,43 @@ pub mod e2e_tests {
 
         Ok(())
     }
+
+    #[ink_e2e::test]
+    async fn test_claim_not_deployer(mut client: ink_e2e::Client<C, E>) -> E2EResult<()> {
```
wojciech-cichocki, line 60: (1) I mean claiming the protocol fee by a non-deployer. Add a test to claim the protocol fee by the changed `fee_receiver`
(2) do not delete this test; instead, change it to claiming the position fee by a non-owner (should fail)

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #118, opened 2023-12-20
Initialized sdk

Sniezka1927:


sdk/src/invariant.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      if (result.isOk && output) {
+        resolve(JSON.parse(output.toString()).ok);
+      } else {
+        reject(new Error(result.asErr.toHuman()?.toString()));
+      }
+    });
+  }
+
+  async changeProtocolFee(fee: { v: number }): Promise<void> {
```
wojciech-cichocki, line 65: I don't like this convention.
1. Instruction should return tx hash
2. Check out this: https://github.com/invariant-labs/psp22-ts/pull/1#discussion_r1432384447

sdk/src/index.ts
```diff
@@ -0,0 +1,25 @@
+import dotenv from "dotenv";
+import { Invariant } from "./invariant.js";
+import { getDeploymentData, initPolkadotJs } from "./utils.js";
+dotenv.config();
+
+(async function main() {
```
wojciech-cichocki, line 6: This convention of top-level function seems to be legacy: `(async function main() {})();`
Try this way: https://github.com/invariant-labs/psp22-ts/blob/main/src/index.ts

sdk/.prettierc.js
```diff
@@ -0,0 +1,16 @@
+const modifiedConfig = {
```
wojciech-cichocki, line 1: I'm not sure but vscode may have problems reading `.prettierc.js` in case there is no on top level of project struct. IMO we can move it to `./` from `./sdk`

wojciech-cichocki (changes requested):

sdk/src/invariant.ts
```diff
@@ -0,0 +1,84 @@
...
+    );
+    this.contract = new ContractPromise(
+      this.api,
+      abi,
+      invariant_deployment.address
+    );
+  }
+
+  async getProtocolFee(): Promise<void> {
+    if (!this.contract) {
```
wojciech-cichocki, line 44: Is there possibility to abstract query similarly to this: https://github.com/invariant-labs/psp22-ts/blob/add-transfer-and-metadata-messages/src/token.ts#L58

sdk/src/invariant.ts
```diff
@@ -0,0 +1,84 @@
...
+    console.log(output?.toHuman());
+
+    if (result.isOk && output) {
+      return JSON.parse(output.toString()).ok;
+    } else {
+      throw new Error(result.asErr.toHuman()?.toString());
+    }
+  }
+
+  async changeProtocolFee(fee: { v: number }): Promise<string> {
```
wojciech-cichocki, line 64: And also here with `sendTx`

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #128, opened 2023-12-28
Add eslint and prettier

zielvna:


sdk/package.json
```diff
@@ -22,5 +22,10 @@
   },
```
wojciech-cichocki, line 22: Add linting as `scripts` in `package.json` also

sdk/.prettierrc
```diff
@@ -0,0 +1,14 @@
+{
+  "arrowParens": "avoid",
```
wojciech-cichocki, line 2: Is there `.prettierrc` file in the root directory of the project already?

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #132, opened 2023-12-28
add psp22 and wrapped azero contracts to sdk and refactor

zielvna:


sdk/src/wrapped_azero.ts
```diff
@@ -0,0 +1,118 @@
...
+      this.storageDepositLimit,
+      0,
+      account,
+      WrappedAZEROTx.Withdraw,
+      [value],
+      this.waitForFinalization
+    )
+  }
+
+  async withdrawWithoutFinalization(account: IKeyringPair, value: number): Promise<string> {
```
wojciech-cichocki, line 82: We should drop approach with separate functions without finalization. Just expose optional `block` param in base function such as `withdraw`

sdk/src/utils.ts
```diff
@@ -1,9 +1,15 @@
...
+import { ContractPromise } from '@polkadot/api-contract'
+import { WeightV2 } from '@polkadot/types/interfaces'
 import { IKeyringPair } from '@polkadot/types/types/interfaces'
 import { getSubstrateChain } from '@scio-labs/use-inkathon/chains'
 import { getBalance, initPolkadotJs as initApi } from '@scio-labs/use-inkathon/helpers'
 import { readFile } from 'fs/promises'
 import { Network } from './network.js'
+import { InvariantQuery, InvariantTx, PSP22Query, PSP22Tx, WrappedAZEROTx } from './schema.js'
+
+export const DEFAULT_REF_TIME = 100000000000
```
wojciech-cichocki, line 11: I wonder if parameters needed for gas limits should be defined per contract, not globally

sdk/src/psp22.ts
```diff
@@ -0,0 +1,199 @@
...
+      this.storageDepositLimit,
+      0,
+      account,
+      PSP22Tx.Transfer,
+      [to, value, data],
+      this.waitForFinalization
+    )
+  }
+
+  async transferWithoutFinalization(
```
wojciech-cichocki, line 103: `Finalization` also here

sdk/src/psp22.ts
```diff
@@ -0,0 +1,199 @@
...
+import { Bytes, Option } from '@polkadot/types'
+import { WeightV2 } from '@polkadot/types/interfaces'
+import { Codec, IKeyringPair } from '@polkadot/types/types'
+import { DeployedContract } from '@scio-labs/use-inkathon'
+import { deployContract } from '@scio-labs/use-inkathon/helpers'
+import { Network } from './network.js'
+import { PSP22Query, PSP22Tx } from './schema.js'
+import { DEFAULT_PROOF_SIZE, DEFAULT_REF_TIME, sendQuery, sendTx } from './utils.js'
+
+export enum PSP22Message {
```
wojciech-cichocki, line 12: Shouldn't this be in the `schema.ts`?

sdk/src/index.ts
```diff
@@ -1,42 +1,108 @@
...
+    symbol,
+    0
+  )
+  await token.load(tokenDeploy.address, tokenData.abi)
+
+  // deploy wrapped azero
+  const wazeroData = await getDeploymentData('wrapped_azero')
+  const wazero = new WrappedAZERO(api, network)
+
+  const wazeroDeploy = await wazero.deploy(account, wazeroData.abi, wazeroData.wasm)
```
wojciech-cichocki, line 65: It's worth using the testnet contract when it exists in the testnet. This emulates the way we will use the SDK on the mainnet or testnet. So, it would be better to just load it on the testnet, instead of deploy and load

wojciech-cichocki (changes requested):

sdk/src/invariant.ts
```diff
@@ -4,105 +4,67 @@ import { WeightV2 } from '@polkadot/types/interfaces'
...
+    proofSize: number = DEFAULT_PROOF_SIZE
   ) {
     this.api = api
-    this.account = account
-    this.weight = api.registry.createType('WeightV2', {
+    this.gasLimit = api.registry.createType('WeightV2', {
       refTime,
       proofSize
     }) as WeightV2
-    this.waitForFinalization = network != Network.Local
```
wojciech-cichocki, line 33: Overall, `waitForFinalization` should remain as is - my mistake. However, if not provided by the user, it fully depends on the network, as here. Then in send tx:
```Typescript
if (result.isCompleted && !waitForFinalization) {
    resolve(result.txHash.toHex();
}

if (result.isFinalized) {
    resolve(result.txHash.toHex();
}
```

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #133, opened 2023-12-28
Add tests

zielvna:


sdk/package.json
```diff
@@ -25,8 +27,12 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
+    "@types/chai": "^4.3.11",
```
wojciech-cichocki, line 30: It looks like some conflict in the IDE
image

sdk/tests/psp22.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      tokenData.wasm,
+      1000,
+      name,
+      symbol,
+      12
+    )
+    await token.load(tokenDeploy.address, tokenData.abi)
+  })
+
+  it('works', async () => {
```
wojciech-cichocki, line 44: If any action can be broken down into smaller steps, I think it's worth doing so.  It can be divided into get metadata, mint, and transfer

sdk/src/index.ts
```diff
@@ -1,43 +0,0 @@
-import { Keyring } from '@polkadot/api'
```
wojciech-cichocki, line 1: We'll ultimately remove this file, but for now, it's more convenient to leave it

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #136, opened 2024-01-02
getting decimal scales

Sniezka1927:


wojciech-cichocki (changes requested):
Try macro with getter

wojciech-cichocki (approved):
Move refactor in separate task

PR #139, opened 2024-01-02
update tests script

zielvna:


sdk/build.sh
```diff
@@ -0,0 +1 @@
+npm run build
```
wojciech-cichocki, line 1: there is no sense to define this kind of script IMO. It will be better to move commands from `tests.sh`:
```
npm i
npm run lint
npm run wasm:build
npm run build
```
and later:
```
# Build and test SDK
cd sdk
./build.sh
./tests.sh
```

wojciech-cichocki (changes requested):

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
wojciech-cichocki, line 1: What's the reason of file?

sdk/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
wojciech-cichocki, line 1:
```
./tests.sh: line 33: ./build.sh: Permission denied
```
IMO it will be also better to add:
```
chmod +x tests.sh build.sh
```
possibility redundant but works on new repo

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #140, opened 2024-01-02
Wasm storage

Sniezka1927:


sdk/math/storage/errors.rs
```diff
@@ -0,0 +1,18 @@
+use crate::alloc::string::ToString;
+use serde::{Deserialize, Serialize};
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub enum InvariantError {
```
wojciech-cichocki, line 8: I think you can already convert all errors

sdk/math/storage/fee_tier.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(Default, Debug, Copy, Clone, PartialEq, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct FeeTier {
+    pub fee: Percentage,
+    pub tick_spacing: u16,
+}
+
+#[wasm_bindgen(js_name = "newFeeTier")]
+pub fn new_fee_tier(js_fee: JsValue, js_tick_spacing: JsValue) -> Result<FeeTier, JsValue> {
```
wojciech-cichocki, line 18: It will be better to use existing logic and the only thing handling in these functions are wrappers. So copy-past constructor from FeeTier and invoke `FeeTier::new()` here. 

It has not significant benefits now, but if we modify it so that the conversion is based on original code, it makes a significant difference.

sdk/math/storage/pool.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub current_tick_index: i32,
+    pub fee_growth_global_x: FeeGrowth,
+    pub fee_growth_global_y: FeeGrowth,
+    pub fee_protocol_token_x: TokenAmount,
+    pub fee_protocol_token_y: TokenAmount,
+    #[tsify(type = "bigint")]
+    pub start_timestamp: u64,
+    #[tsify(type = "bigint")]
+    pub last_timestamp: u64,
+    pub fee_receiver: String,
```
wojciech-cichocki, line 26: Is there possibility to use the same struct. but changed types using `tsify` or other solution?

sdk/package-lock.json
```diff
@@ -5208,6 +5208,11 @@
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
       }
+    },
```
wojciech-cichocki, line 5211: do you need this now?

wojciech-cichocki (changes requested):

sdk/math/storage/fee_tier.rs
```diff
@@ -0,0 +1,36 @@
...
+use decimal::*;
+use serde::{Deserialize, Serialize};
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+#[derive(Default, Debug, Copy, Clone, PartialEq, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct FeeTier {
+    pub fee: Percentage,
+    pub tick_spacing: u16,
```
wojciech-cichocki, line 14: We convert all int/uint values to `bigInt`, avoiding the conversion of integral values to ts `number` type

wojciech-cichocki (changes requested):

sdk/src/index.ts
```diff
@@ -32,6 +41,16 @@ const main = async () => {
     const scales = getDecimalScales()
     console.log(scales)
   }
+  {
+    const feeTier: FeeTier = newFeeTier({ v: 10n }, 55)
```
wojciech-cichocki, line 45: what about literal of `bigint` here

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #147, opened 2024-01-03
add get tick and is tick initalized tests to sdk

zielvna:


sdk/src/invariant.ts
```diff
@@ -265,15 +265,34 @@ export class Invariant {
...
+          liquidityGross: { v: BigInt(result.ok.liquidityGross.v) },
+          sqrtPrice: { v: BigInt(result.ok.sqrtPrice.v) },
+          feeGrowthOutsideX: { v: BigInt(result.ok.feeGrowthOutsideX.v) },
+          feeGrowthOutsideY: { v: BigInt(result.ok.feeGrowthOutsideY.v) },
+          secondsOutside: BigInt(result.ok.secondsOutside)
+        }
+      }
+    } else {
+      return {
+        err: result.err
```
wojciech-cichocki, line 293: Better will be to return an enum. These errors are our `InvarinatError` defined in contract. Conversion should be quite easy and fully generic.
1. `InvarinatError` enum is in wasm
2. The returned error is the name of the enum as string, similar to the case in the `Network` enum and the `valueOf` function

wojciech-cichocki (changes requested):

sdk/tests/invariant.test.ts
```diff
@@ -73,6 +73,61 @@ describe('invariant', async () => {
...
+      -10n,
+      10n,
+      { v: 1000000n },
+      { v: 0n },
+      { v: 100000000000000000000000000n }
+    )
+
+    const lowerTick = await invariant.getTick(account, poolKey, -10n)
+    assert.deepEqual(lowerTick.index, -10n)
+    await catchError(invariant.getTick(account, poolKey, 0n), 'TickNotFound')
```
wojciech-cichocki, line 119: change error to enum type

sdk/tests/invariant.test.ts
```diff
@@ -73,6 +73,61 @@ describe('invariant', async () => {
...
+    const upperTick = await invariant.getTick(account, poolKey, 10n)
+    assert.deepEqual(upperTick.index, 10n)
```
wojciech-cichocki, line 121: what about asserting other tick's fields?

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #150, opened 2024-01-04
Event handling

zielvna:


sdk/src/invariant.ts
```diff
@@ -60,6 +65,45 @@ export class Invariant {
 
   async load(deploymentAddress: string, abi: any): Promise<void> {
     this.contract = new ContractPromise(this.api, abi, deploymentAddress)
+    this.abi = new Abi(abi)
+
+    this.api.query.system.events((events: any) => {
```
wojciech-cichocki, line 70: Well, I'm not sure if it's the correct approach. Let's summarize: we need two types of events:
1. Events returned from the currently sent transaction by the user. Have you checked [this](https://polkadot.js.org/docs/api-contract/start/contract.tx#events)?
2. Events that we can subscribe to for all contract events. To illustrate, imagine subscribing to certain events, and when a user sends a transaction using another client, after finalization, you should receive emitted events from that transaction.

Two extra things:
1. You can also investing if is the possibility to subscribe change of contract storage.
2. Events are commonly used for calculating statistics. Could you please check if it's possible to fetch past events?

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

wojciech-cichocki (approved):

PR #151, opened 2024-01-04
E2e position

Sniezka1927:


sdk/tests/invariant.test.ts
```diff
@@ -248,4 +248,198 @@ describe('invariant', async () => {
...
+      poolKey,
+      lowerTickIndex,
+      upperTickIndex,
+      { v: 1000000n },
+      { v: 0n },
+      { v: 100000000000000000000000000n }
+    )
+    const position = await invariant.getPosition(account, 0n)
+    assert.deepEqual(position, {
+      poolKey: position.poolKey,
```
wojciech-cichocki, line 291: `position.poolKey.equals(position.poolKey)` will always be true (regardless of type). I don't see a reason to compare like this. It would be better to compare it with the `poolKey` object defined above.

sdk/tests/invariant.test.ts
```diff
@@ -248,4 +248,198 @@ describe('invariant', async () => {
...
+      liquidity: { v: 1000000n },
+      lowerTickIndex: lowerTickIndex,
```
wojciech-cichocki, line 293: Here's example of better comparison

sdk/tests/invariant.test.ts
```diff
@@ -248,4 +248,198 @@ describe('invariant', async () => {
...
+      upperTickIndex: upperTickIndex,
+      feeGrowthInsideX: { v: 0n },
+      feeGrowthInsideY: { v: 0n },
+      lastBlockNumber: position.lastBlockNumber,
```
wojciech-cichocki, line 297: Same example of: `position.lastBlockNumber === position.lastBlockNumber`. It's impossible to predict the value of this field. I suggest defining test-specific utility file. Add there `positionEquals` function that compare fields as skip comparing fields like the one mentioned

sdk/tests/invariant.test.ts
```diff
@@ -248,4 +248,198 @@ describe('invariant', async () => {
...
+      tokensOwedX: 0n,
+      tokensOwedY: 0n
+    })
+  })
+  it('remove position', async () => {
+    if (!token0.contract?.address || !token1.contract?.address || !invariant.contract?.address) {
```
wojciech-cichocki, line 303: Well each test starts with opening position. You can easily utilize this and add this opening of a position using the `beforeEach` hook. Create a nest in the test structure like this:
```
describe 'invarinat'
  beforeEach (deploy invariant contract)
  describe 'positions'
    beforeEach (open position)
    it 'create position'
    it 'remove position'
    ...
```
The structure will be clearer and it will be easier to divide these tests between files

wojciech-cichocki (changes requested):

sdk/tests/invariant.test.ts
```diff
@@ -248,4 +269,240 @@ describe('invariant', async () => {
     const pools = await invariant.getPools(account)
     assert.deepEqual(pools.length, 1)
   })
+  describe('positions', async () => {
+    const lowerTickIndex = -20n
+    const upperTickIndex = 10n
```
wojciech-cichocki, line 274: I like the approach of defining all common configuration constants here. You can also consider moving constants like this:
```ts
const feeTier = newFeeTier({ v: 6000000000n }, 10)
```

sdk/tests/invariant.test.ts
```diff
@@ -248,4 +269,240 @@ describe('invariant', async () => {
...
+        )
+
+        const position = await invariant.getPosition(account, account.address, 0n)
+        const pool = await invariant.getPool(
+          account,
+          token0.contract.address.toString(),
+          token1.contract.address.toString(),
+          feeTier
+        )
+        const expectedTokensClaimed = BigInt(5)
```
wojciech-cichocki, line 498: use literals `5n`

sdk/tests/invariant.test.ts
```diff
@@ -248,4 +269,240 @@ describe('invariant', async () => {
...
+        upperTickIndex: upperTickIndex,
+        feeGrowthInsideX: { v: 0n },
+        feeGrowthInsideY: { v: 0n },
+        lastBlockNumber: 0n,
+        tokensOwedX: 0n,
+        tokensOwedY: 0n
+      }
+      await positionEquals(position, expectedPosition)
+    })
+    it('remove position', async () => {
```
wojciech-cichocki, line 330: There is a lack of 'create position' entry. I'm guessing it is fully inside 'before'. You can transfer the getPosition and assert inside 'create position'. There is no need to repeat the same assertion inside all tests

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #152, opened 2024-01-04
Refactor getting positions

Sniezka1927:


src/test_helpers/entrypoints.rs
```diff
@@ -243,9 +243,9 @@ macro_rules! transfer_position {
 
 #[macro_export]
 macro_rules! get_position {
-    ($client:ident, $dex:ty, $dex_address:expr, $index:expr, $caller:ident) => {{
+    ($client:ident, $dex:ty, $dex_address:expr, $owner:expr, $index:expr, $caller:ident) => {{
```
wojciech-cichocki, line 246: All tests were written with the idea of calling the position for the invoking user. I wouldn't change this approach. It's better to take the caller at the macro level and pass it to the entry point without modifying any test

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #154, opened 2024-01-08
Add private constructor

Sniezka1927:


sdk/src/invariant.ts
```diff
@@ -50,6 +53,36 @@ export class Invariant {
     }) as WeightV2
     this.storageDepositLimit = storageDepositLimit
     this.waitForFinalization = network != Network.Local
+    this.contract = new ContractPromise(this.api, abi, deploymentAddress)
+  }
+
+  static async create(
```
wojciech-cichocki, line 59: There is no need to create new function `create`. Now, `deploy` and `load` have lost purpose. `deploy` and `load` should be static methods also.

Yes, you can create a function `create`, but it should be called underneath the `deploy` and `load` calls. Importantly, there should be a separate function for just loading an existing contract, used in testnet and development, and another for deploying and loading a contract used in a localnet.

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):
Now there is also no need to validate if the contract is `null` in other spaces such as `sendTx` and `sendQuery`.

wojciech-cichocki (approved):

PR #158, opened 2024-01-08
Expose macro

Sniezka1927:


sdk/math/math.rs
```diff
@@ -17,14 +20,16 @@ pub struct SingleTokenLiquidity {
     pub amount: TokenAmount,
 }
 
-#[allow(dead_code)]
+#[wasm_wrapper]
 pub fn get_liquidity_by_x(
     x: TokenAmount,
-    lower_tick: i32,
-    upper_tick: i32,
+    lower_tick: i64,
```
wojciech-cichocki, line 26: It would be preferred not to change the function's interface. Such as here: https://github.com/invariant-labs/protocol-a0/pull/158/files#diff-7865676860776782b480153cac779a33422a5f9ffce1134dc0458ebe2aee4008L270

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #160, opened 2024-01-09
Add events to sdk tests

zielvna:


sdk/tests/invariant.test.ts
```diff
@@ -245,13 +252,162 @@ describe('invariant', async () => {
...
+        account,
+        poolKey,
+        -10n,
+        10n,
+        { v: 10000000000000n },
+        { v: 1000000000000000000000000n },
+        { v: 1000000000000000000000000n }
+      )
+
+      assert.deepEqual((result.events[0] as any).address, account.address.toString())
```
wojciech-cichocki, line 288: Is there a result that returns a contract when you call the entry point always the last event in the array? It seems it should be the last. If so, we can utilize this fact and return strongly typed events

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):
Not sure if the events struct from the tx was exposed in the math module by `wasm_bindgen`. If not, you need to export it

PR #164, opened 2024-01-11
add get all ticks message

zielvna:

PR #165, opened 2024-01-11
Add get tickmap

Sniezka1927:


src/lib.rs
```diff
@@ -941,6 +943,36 @@ pub mod invariant {
...
             self.fee_tiers.get_all()
         }
+
+        #[ink(message)]
+        fn get_tickmap(
+            &self,
+            pool_key: PoolKey,
+            starting_chunk: u16,
+            finishing_chunk: u16,
+        ) -> Vec<(u16, u64)> {
```
wojciech-cichocki, line 953: There is no need for pagination here. We need the tickmap only to simulate swaps. In the interface, you should pass center tick index. It should return the maximum number of ticks symmetrically in the current chunk; half below and half above. If there are not enough ticks on one side, get more from the opposite side

wojciech-cichocki (changes requested):

src/lib.rs
```diff
@@ -941,6 +943,60 @@ pub mod invariant {
...
+            let max_chunks_queried = 2047;
+            while tickmap_slice.len() < max_chunks_queried {
+                if current_chunk_index + step > MAX_CHUNK {
+                    let chunk = self
+                        .tickmap
+                        .bitmap
+                        .get((current_chunk_index - step, pool_key))
+                        .unwrap_or(0);
+                    step += 1;
+                    tickmap_slice.insert(0, chunk);
```
wojciech-cichocki, line 970: make sure about the order of these chunks. It should be the same order as in the tickmap state (lower index in tickmap means a lower index in the result vector)

src/e2e/get_tickmap.rs
```diff
@@ -0,0 +1,230 @@
...
+        let tickmap = get_tickmap!(
+            client,
+            InvariantRef,
+            dex,
+            pool_key,
+            pool.current_tick_index,
+            alice
+        );
+
+        assert_eq!(tickmap.len(), 2047);
```
wojciech-cichocki, line 99: There is definitely a need to check the result of the tickmap, not just the length of this tickmap

src/e2e/get_tickmap.rs
```diff
@@ -0,0 +1,230 @@
...
```
wojciech-cichocki, line 227: Same thing, you need to validate the content of tickmap in all `get_tickmap` tests

src/e2e/get_tickmap.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        Ok(())
+    }
+
+    #[ink_e2e::test]
+    async fn test_get_tickmap_close_to_bottom_limit(
```
wojciech-cichocki, line 169: Also, add a test that checks whether the returned chunks are symmetrical. I mean case when there are many more ticks on one side

sdk/src/invariant.ts
```diff
@@ -566,4 +566,19 @@ export class Invariant {
       block
     ) as Promise<SwapRouteTxResult>
   }
+  async getTickmap(
+    account: IKeyringPair,
+    poolKey: PoolKey,
+    currentTickIndex: bigint
+  ): Promise<bigint[][]> {
+    const result = (await sendQuery(
```
wojciech-cichocki, line 574: That's not sufficient.
1. Create entire tickmap (create tickmap with all 0, then update initalized tick)
2. You can test u64 type here (will be better than bigint)

sdk/tests/get_tickmap.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      account,
+      poolKey,
+      ticks[2],
+      ticks[3],
+      { v: 10n },
+      pool.sqrtPrice,
+      pool.sqrtPrice
+    )
+
+    const tickmap = await invariant.getTickmap(account, poolKey, pool.currentTickIndex)
```
wojciech-cichocki, line 65: You should validate the created tickmap, not only the result from `getTickmap`

wojciech-cichocki (changes requested):

sdk/tests/get_tickmap.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    assert.deepEqual(tickmap[3465], 9223372036854775809n)
```
wojciech-cichocki, line 70: And all remaining chunks should be `0`, right? It's worth validating this. It could be helpful to avoid possible hidden errors

sdk/src/utils.ts
```diff
@@ -302,3 +303,15 @@ const isArray = (value: any): boolean => {
...
+export const constructTickmap = (tickmap: bigint[][]): bigint[] => {
+  const maxChunk = getMaxChunk()
+  const tmpTickmap = []
+  for (let i = 0; i <= maxChunk; i++) {
+    tmpTickmap[i] = 0n
+  }
+  for (let i = 0; i < tickmap.length; i++) {
+    tmpTickmap[Number(tickmap[i][0])] = tickmap[i][1]
+  }
+  return tmpTickmap
```
wojciech-cichocki, line 316: Is this really `tmpTickmap`? It seems to be the result, so I suggest changing the naming, perhaps to `tickmap` and the parameter of the function as `initializedChunks`

sdk/src/utils.ts
```diff
@@ -302,3 +303,15 @@ const isArray = (value: any): boolean => {
...
   return typeof value === 'object' && value !== null
 }
+
+export const constructTickmap = (tickmap: bigint[][]): bigint[] => {
+  const maxChunk = getMaxChunk()
+  const tmpTickmap = []
+  for (let i = 0; i <= maxChunk; i++) {
+    tmpTickmap[i] = 0n
+  }
+  for (let i = 0; i < tickmap.length; i++) {
```
wojciech-cichocki, line 313: I suggest updating to slightly more readable loops, like:
```Typescript
  const tmpTickmap = new Array<bigint>(maxChunk + 1).fill(0n);

  for (const [chunkIndex, value] of tickmap) {
    tmpTickmap[Number(chunkIndex)] = value;
  }
```

wojciech-cichocki (changes requested):

src/e2e/get_tickmap.rs
```diff
@@ -0,0 +1,493 @@
...
+            init_tick,
+            alice
+        );
+        assert!(result.is_ok());
+
+        let pool = get_pool!(client, InvariantRef, dex, token_x, token_y, fee_tier).unwrap();
+
+        let liquidity_delta = Liquidity::new(1000);
+
+        create_position!(
```
wojciech-cichocki, line 134: Create another position here with ticks: `<min_tick, max_tick>`. Then, add an assertion to ensure of multiple edges with `max_chunk` etc.

src/e2e/get_tickmap.rs
```diff
@@ -0,0 +1,493 @@
...
+            pool_key,
+            pool.current_tick_index,
+            alice
+        );
+
+        assert_eq!(tickmap[0], (0, 0b11));
+        assert_eq!(
+            tickmap[1],
+            (
+                6931,
```
wojciech-cichocki, line 243: Here, you can easily validate the `get_max_chunk` function. The index returned by `get_tickmap` should match the invocation of `get_max_chunk`.

src/e2e/get_tickmap.rs
```diff
@@ -0,0 +1,493 @@
...
+            init_tick,
+            alice
+        );
+        assert!(result.is_ok());
+
+        let pool = get_pool!(client, InvariantRef, dex, token_x, token_y, fee_tier).unwrap();
+
+        let liquidity_delta = Liquidity::new(1000);
+
+        for i in (6..10048).step_by(64) {
```
wojciech-cichocki, line 322: Is this more than `MAX_CHUNK_AMOUNT/2 ==  819`? it will correctly test whether if there are no ticks on one side, more can be taken from the other

sdk/tests/get_tickmap.test.ts
```diff
@@ -0,0 +1,173 @@
...
+      10n,
+      pool.sqrtPrice,
+      pool.sqrtPrice
+    )
+
+    const tickmap = await invariant.getTickmap(account, poolKey, pool.currentTickIndex)
+    assert.deepEqual(tickmap[3465], 9223372036854775809n)
+    for (const [chunkIndex, value] of tickmap.entries()) {
+      if (chunkIndex === 3465) {
+        assert.deepEqual(value, 9223372036854775809n)
```
wojciech-cichocki, line 63: you can also use binary or hex notation: `0x1ABCn` nad `0b10101n`

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):
Resolve conflicts. Also replace this default `Number` casting to `safeIntgerCasting`:
https://github.com/invariant-labs/protocol-a0/pull/164/commits/48d11fab1c10cc31810808710566ae59ae00cbe3#diff-1f487bce7f721a927d8d0190259b8cff4e25b1d36f3ec8b0ac64686e93c8479fR86

wojciech-cichocki (approved):

PR #166, opened 2024-01-12
create calculate price impact function to utils and tests

zielvna:


sdk/src/utils.ts
```diff
@@ -271,3 +271,22 @@ export const getDeploymentData = async (
...
   }
 }
+
+export const DENOMINATOR = BigInt(Math.pow(10, 12))
+
+export const calculatePriceImpact = (
+  startingSqrtPrice: SqrtPrice,
+  endingSqrtPrice: SqrtPrice
+): Percentage => {
+  const startingPrice = (startingSqrtPrice.v as bigint) * (startingSqrtPrice.v as bigint)
```
wojciech-cichocki, line 281: Why here's casting to `bigint`? We pay lot of attention to current types to avoid thinking about it. If it's intentional redundancy, I'm not really supporter of this approach

sdk/src/utils.ts
```diff
@@ -271,3 +271,22 @@ export const getDeploymentData = async (
     throw new Error(`${contractName}.json or ${contractName}.wasm not found`)
   }
 }
+
+export const DENOMINATOR = BigInt(Math.pow(10, 12))
+
+export const calculatePriceImpact = (
```
wojciech-cichocki, line 277: This code is from prehistory of Invarinat. Wrtie this function in good style. Here's formula:
image

sdk/tests/utils.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe('utils', () => {
+  describe('test calculatePriceImpact', () => {
+    it('increasing price', () => {
+      // price change       120 -> 599
+      // real price impact  79.96661101836...%
+      const startingSqrtPrice = { v: 10954451150103322269139395n }
+      const endingSqrtPrice = { v: 24474476501040834315678144n }
+      const priceImpact = calculatePriceImpact(startingSqrtPrice, endingSqrtPrice)
+      assert.equal(priceImpact.v as bigint, 799666110184n)
```
wojciech-cichocki, line 12: Is this casting to `bigint` inside both tests necessary?

wojciech-cichocki (changes requested):

sdk/src/utils.ts
```diff
@@ -271,3 +271,22 @@ export const getDeploymentData = async (
     throw new Error(`${contractName}.json or ${contractName}.wasm not found`)
   }
 }
+
+export const DENOMINATOR = BigInt(Math.pow(10, 12))
```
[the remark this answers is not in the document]
wojciech-cichocki: Try this solution: https://github.com/invariant-labs/protocol-a0/pull/167#discussion_r1451898576

sdk/src/utils.ts
```diff
@@ -226,6 +228,23 @@ export const getDeploymentData = async (
...
 
+export const calculatePriceImpact = (
+  startingSqrtPrice: SqrtPrice,
+  endingSqrtPrice: SqrtPrice
+): Percentage => {
+  const startingPrice = startingSqrtPrice.v * startingSqrtPrice.v
+  const endingPrice = endingSqrtPrice.v * endingSqrtPrice.v
+
+  const nominator = BigInt(Math.abs(Number(startingPrice - endingPrice)))
+  const denominator = BigInt(Math.max(Number(startingPrice), Number(endingPrice)))
```
wojciech-cichocki, line 239: What about domain and limit `number.max_safe_integer`? IMO there will be safety to use native `bigint`. So:
max: `a > b ? a : b`
abs: `x < 0n ? -x : x`

sdk/src/utils.ts
```diff
@@ -226,6 +228,23 @@ export const getDeploymentData = async (
...
+
+  return {
+    v: BigInt(
+      Math.round((Number(nominator) / Number(denominator)) * Number(getPercentageDenominator()))
```
wojciech-cichocki, line 243: Check if domain is safe. If not, use native `bigint`. Math.round you can easily replace will `mul_up`

wojciech-cichocki (approved):

PR #167, opened 2024-01-12
create calculate price after slippage to utils and tests

zielvna:


sdk/src/utils.ts
```diff
@@ -271,3 +271,20 @@ export const getDeploymentData = async (
     throw new Error(`${contractName}.json or ${contractName}.wasm not found`)
   }
 }
+
+export const DENOMINATOR = BigInt(Math.pow(10, 12))
+export const PRICE_DENOMINATOR = BigInt(Math.pow(10, 24))
```
wojciech-cichocki, line 276: Best solution will be to extend scale macro: https://github.com/invariant-labs/protocol-a0/blob/master/sdk/math/wasm_helpers.rs#L56
Add denominator similarity as was added `get${decimal}Scale`.

wojciech-cichocki (changes requested):

sdk/src/testUtils.ts
```diff
@@ -46,3 +47,15 @@ export const getEnvTestAccount = async (keyring: Keyring): Promise<IKeyringPair>
 
   return keyring.addFromUri(accountUri)
 }
+
+export const toSqrtPrice = (x: bigint, decimals: bigint = 0n): SqrtPrice => {
```
wojciech-cichocki, line 51: Well, we need to repeat this for all types. There is already equivalent of this function in rust `from_decimal`. Please add this function in wasm scale macro. Same as here: https://github.com/invariant-labs/protocol-a0/pull/167/files#r1451898576

wojciech-cichocki (changes requested):

sdk/tests/utils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      const expected = { v: 990000000000756263920003n }
+
+      const limitSqrt = calculatePriceAfterSlippage(price, slippage, false)
+      const limit = ((limitSqrt.v as bigint) * (limitSqrt.v as bigint)) / PRICE_DENOMINATOR
+
+      assert.equal(limit, expected.v)
+    })
+
+    it('slippage of 0.5% up', async () => {
```
wojciech-cichocki, line 55: Defining a function in a test as `async` seems unnecessary. There are no asycn function invocations here

wojciech-cichocki (changes requested):

sdk/math/wasm_helpers.rs
```diff
@@ -62,14 +63,28 @@ extern "C" {
...
             #[wasm_bindgen]
             #[allow(non_snake_case)]
             pub fn [<get $decimal Scale >] () -> BigInt {
                 BigInt::from($decimal::scale())
             }
+
+            #[wasm_bindgen]
+            #[allow(non_snake_case)]
+            pub fn [<get $decimal Denominator >] () -> BigInt {
+                BigInt::from(10u128.pow($decimal::scale() as u32))
```
wojciech-cichocki, line 78: There is already denominator implemented in decimal: `BigInt::from($decimal::from_integer(1).get())`

sdk/src/utils.ts
```diff
@@ -226,6 +228,37 @@ export const getDeploymentData = async (
...
+
+export const calculateSqrtPriceAfterSlippage = (
+  sqrtPrice: SqrtPrice,
+  slippage: Percentage,
+  up: boolean
+): SqrtPrice => {
+  const multiplier = getPercentageDenominator() + (up ? slippage.v : -slippage.v)
+  const multiplierSqrt = sqrt(multiplier * getPercentageDenominator())
+
+  return { v: (sqrtPrice.v * multiplierSqrt) / getPercentageDenominator() }
```
wojciech-cichocki, line 259: The issue is that you always receive the sqrt price of the result with precision of `10^12`, causing the last 12 digits to be consistently zeros. We should change the scale to `24` (from percentage scale to sqrtPrice scale)

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #168, opened 2024-01-12
calculating token amounts from position liquidity

Sniezka1927:


sdk/math/wasm_helpers.rs
```diff
@@ -81,3 +90,39 @@ macro_rules! resolve {
         }
     }};
 }
+
+#[wasm_bindgen(js_name = "calculateTokenAmountsFromPosition")]
+pub fn calculate_token_amounts_from_position_liquidity(
```
wojciech-cichocki, line 95: I was wondering whether such functions should be defined in TS or Rust, and we should create function in TS. Inside call `calculateAmountDelta` wasm function.

Why use TS?
1. There is no duplicating math logic here (so no risk of inconsistency)
2. This function doesn't calculate anything, it's just wrapping the interface. In TS it can be done much faster and we can expect the same result.
3. From the sdk client's perspective, we should build easy to use and high-level interface.
E.g. this interface looks much better:
```Typescript
const { x, y } = calculateTokenAmountsFromPosition(pool, position)
```
than this one:
```Typescript
const { x, y } = calculateTokenAmountsFromPosition(
  pool.currentTickIndex,
  pool.sqrtPrice,
  position.liquidity,
  position.upperTickIndex,
  position.lowerTickIndex,
  position.tokensOwedX,
  position.tokensOwedY
)
```

sdk/tests/invariant.test.ts
```diff
@@ -449,6 +451,36 @@ describe('invariant', async () => {
...
+      const { x, y } = calculateTokenAmountsFromPosition(
+        pool.currentTickIndex,
+        pool.sqrtPrice,
+        position.liquidity,
+        position.upperTickIndex,
+        position.lowerTickIndex,
+        position.tokensOwedX,
+        position.tokensOwedY
+      )
+      assert.deepEqual(x, providedAmount - 1n)
```
wojciech-cichocki, line 481: You can add comment that `1n` results from different rounding

sdk/tests/invariant.test.ts
```diff
@@ -449,6 +451,36 @@ describe('invariant', async () => {
...
+      const position = await invariant.getPosition(account, account.address, 0n)
+      const pool = await invariant.getPool(
+        account,
+        token0.contract.address.toString(),
+        token1.contract.address.toString(),
+        feeTier
+      )
+
+      const providedAmount = 500n
+      const expectedYAmount = getLiquidityByX(
```
wojciech-cichocki, line 464: I don't like this syntax, I think it is much less readable:
```TypeScript
const expectedYAmount = getLiquidityByX(...).amount
```
IMO this one is better
```TypeScript
const { amount: expectedYAmount } = getLiquidityByX(...)
```
1. It's easy to miss reference to field.
2. harder to take another result parameter

wojciech-cichocki (changes requested):

sdk/src/utils.ts
```diff
@@ -216,6 +225,15 @@ export const getDeploymentData = async (
   }
 }
 
+export const calculateTokensAmountFromPositionLiquidity = (pool: Pool, position: Position) => {
```
wojciech-cichocki, line 228: Add type support here. Now it looks like this:
```Typescript
const calculateTokensAmountFromPositionLiquidity: (pool: Pool, position: Position) => any
```

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #174, opened 2024-01-16
added static load method for psp22

Sniezka1927:


wojciech-cichocki (changes requested):
Currently, each instance of the `PSP22` class is associated with a specific psp22 token. The idea was for one instance of the `PSP22` class not to be tied to a specific token/contract and to have the ability to call multiple. But in general, wait for the merging of this https://github.com/invariant-labs/protocol-a0/pull/175 into the master before making changes

sdk/src/psp22.ts
```diff
@@ -90,7 +93,19 @@ export class PSP22 {
     )
   }
 
-  async mint(account: IKeyringPair, value: bigint, block: boolean = true): Promise<TxResult> {
+  async load(address: string) {
```
wojciech-cichocki, line 96: `load` naming creates collisions with an already existing loading. To make it clear I suggest change naming. Maybe: `changeContractAddress`

sdk/src/psp22.ts
```diff
@@ -90,7 +93,19 @@ export class PSP22 {
...
+    this.contract = new ContractPromise(this.api, this.abi, address)
+  }
+
+  async mint(
+    account: IKeyringPair,
+    value: bigint,
+    address: string = this.contract.address.toString(),
+    block: boolean = true
+  ): Promise<TxResult> {
+    if (!isLoaded(address, this.contract.address.toString())) {
```
wojciech-cichocki, line 106: It's not the best coding practice. When adding any new method, you must remember to check the load. This makes it harder to extend the code and easy to create bugs when you forget to load.

So, all in all, it will be better to add `setContractAddress`, and the client will change it manually before call

sdk/src/utils.ts
```diff
@@ -302,3 +302,7 @@ const isArray = (value: any): boolean => {
 const isObject = (value: any): boolean => {
   return typeof value === 'object' && value !== null
 }
+
+export const isLoaded = (providedAddress: string, loadedAddress: string): boolean => {
```
wojciech-cichocki, line 306: You probably won't need this function after changes

sdk/tests/invariant.test.ts
```diff
@@ -96,8 +96,18 @@ describe('invariant', async () => {
       0n
     )
 
-    await token0.approve(account, invariant.contract.address.toString(), 1000000000n)
-    await token1.approve(account, invariant.contract.address.toString(), 1000000000n)
+    await token0.approve(
```
wojciech-cichocki, line 99: You should use only one instance of `PSP22` class. This class is no longer associated permanently with specific token

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):

PR #175, opened 2024-01-16
refactor deploy and load of contracts in sdk

zielvna:


sdk/src/invariant.ts
```diff
@@ -53,72 +53,87 @@ export class Invariant {
...
   static async getContract(
     api: ApiPromise,
-    account: IKeyringPair,
-    storageDepositLimit: number | null = null,
-    refTime: number = DEFAULT_REF_TIME,
-    proofSize: number = DEFAULT_PROOF_SIZE,
-    initFee: Percentage,
-    network: Network
+    network: Network,
+    account?: IKeyringPair,
```
wojciech-cichocki, line 76: Name it `deployer: IKeyringPair`. From the interface perspective, you don't know what type of `account` it is. It might be confused with an account meant to sign the transaction, as that is also named `account`.

wojciech-cichocki: do the same in `deploy` function

sdk/src/invariant.ts
```diff
@@ -53,72 +53,87 @@ export class Invariant {
...
+      network,
+      deploymentData.abi,
+      deploy.address,
+      options?.storageDepositLimit,
+      options?.refTime,
+      options?.proofSize
+    )
+  }
+
+  static async load(
```
wojciech-cichocki, line 120: Mark all functions that you assume should not be directly called by the sdk client as `private`. It appears that `load` is one such function

wojciech-cichocki (changes requested):

sdk/src/utils.ts
```diff
@@ -174,15 +174,7 @@ export const deployInvariant = async (
...
-  return Invariant.getContract(
-    api,
-    account,
-    null,
-    DEFAULT_REF_TIME,
-    DEFAULT_PROOF_SIZE,
-    initFee,
-    network
-  )
+  return Invariant.getContract(api, network, account, initFee)
```
wojciech-cichocki, line 177: I'm wondering about the purpose of these functions. All they do is copy the parameters for the `getContract` call. There are two solutions. Choose the better one:
1. Since all `deploy[contract]` functions are used only for tests, and for all tests, we use the `localnet`, you can remove the `Network` parameter and pass `Network.Local` instead.
2. Delete `deploy[contract]` and use `getContract`.

sdk/src/utils.ts
```diff
@@ -194,26 +186,15 @@ export const deployPSP22 = async (
...
+  return PSP22.getContract(api, network, account, supply, name, symbol, decimals)
 }
 
 export const deployWrappedAZERO = async (
   api: ApiPromise,
   account: IKeyringPair,
   network: Network
 ): Promise<WrappedAZERO> => {
-  return WrappedAZERO.getContract(api, account, null, DEFAULT_REF_TIME, DEFAULT_PROOF_SIZE, network)
+  return WrappedAZERO.getContract(api, network, account, process.env.WAZERO_ADDRESS)
```
wojciech-cichocki, line 197: It's probably mistake to include both the deployer and contract address in the get contract call

sdk/src/wrapped_azero.ts
```diff
@@ -2,10 +2,9 @@ import { ApiPromise } from '@polkadot/api'
 import { ContractPromise } from '@polkadot/api-contract'
```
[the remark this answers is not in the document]
wojciech-cichocki: This should be also apply to `sdk/tests/` dir

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #180, opened 2024-01-22
add sqrt price to price and vice versa conversion and tests

zielvna:


sdk/src/utils.ts
```diff
@@ -299,3 +299,11 @@ const isArray = (value: any): boolean => {
 const isObject = (value: any): boolean => {
   return typeof value === 'object' && value !== null
 }
+
+export const sqrtPriceToPrice = (sqrtPrice: SqrtPrice): bigint => {
```
wojciech-cichocki, line 303: Add the type `Price`. This type will help to avoid mistakes when passing `Price` and `SqrtPrice`

wojciech-cichocki (changes requested):

PR #181, opened 2024-01-22
add parse liquidity on ticks function

zielvna:


sdk/src/utils.ts
```diff
@@ -308,3 +309,20 @@ export const sqrtPriceToPrice = (sqrtPrice: SqrtPrice): Price => {
 export const priceToSqrtPrice = (price: Price): SqrtPrice => {
   return sqrt(price * getSqrtPriceDenominator())
 }
+
+interface LiquidityParsedTick {
+  liquidity: Liquidity
+  index: bigint
+}
+
+export const parseLiquidityOnTicks = (ticks: Tick[]): LiquidityParsedTick[] => {
```
wojciech-cichocki, line 318: use `LiquidityTick | Tick` type

sdk/src/utils.ts
```diff
@@ -308,3 +309,20 @@ export const sqrtPriceToPrice = (sqrtPrice: SqrtPrice): Price => {
 export const priceToSqrtPrice = (price: Price): SqrtPrice => {
   return sqrt(price * getSqrtPriceDenominator())
 }
+
+interface LiquidityParsedTick {
```
wojciech-cichocki, line 313: `LiquidityBreakpoint` is more accurate name for what this struct represents

sdk/src/utils.ts
wojciech-cichocki, line 318: according to name in struct name you can rename this function: `calculateLiquidityBreakpoints`

wojciech-cichocki (changes requested):

PR #187, opened 2024-01-24
refactor sdk and update index file

zielvna:


sdk/src/schema.ts
```diff
@@ -101,3 +101,8 @@ export type ContractOptions = {
   refTime: number
   proofSize: number
 }
+
+export interface Result<T, E> {
```
wojciech-cichocki, line 105: There was discussion about this topic before. All in all, we decided to follow the TypeScript convention for handling errors. Therefore:
1. Always return an "ok" result without any wrapper.
2. Do not add `Error` to the function signature.
3. Instead of returning an error, throw it, and then use the catch statement

Based on these decisions, this interface should not be used

wojciech-cichocki (changes requested):

sdk/src/invariant.ts
```diff
@@ -507,7 +499,7 @@ export class Invariant {
     amount: TokenAmount,
     byAmountIn: boolean,
     sqrtPriceLimit: SqrtPrice
-  ): Promise<QuoteResult> {
+  ): Promise<Result<QuoteResult, InvariantError>> {
```
wojciech-cichocki, line 502: https://github.com/invariant-labs/protocol-a0/pull/187/files#r1465819320

sdk/src/index.ts
```diff
@@ -1,177 +1,164 @@
...
+  const api = await initPolkadotApi(Network.Local) // initialize api, use enum to specify the network
 
-    const deltaYUp = getDeltaY(sqrtPriceA, sqrtPriceB, liquidity, true)
-    const deltaYDown = getDeltaY(sqrtPriceA, sqrtPriceB, liquidity, false)
-    console.log(deltaYUp)
-    console.log(deltaYDown)
-  }
+  // initialize account, you can use your own wallet by pasting mnemonic phase
+  const keyring = new Keyring({ type: 'sr25519' })
+  const account = await keyring.addFromUri('//Alice')
```
wojciech-cichocki, line 23: Here, `await` is not required as `addFromUri` isn't an `async` function

sdk/src/index.ts
```diff
@@ -1,177 +1,164 @@
...
-  console.log(
-    'test account balance: ',
-    (await getBalance(api, testAccount.address)).balanceFormatted
-  )
+  // approve token x transfer
+  await psp22.setContractAddress(poolKey.tokenX)
+  await psp22.approve(account, invariant.contract.address.toString(), amount)
+
+  // get estimated result of swap
+  const quoteResult = await invariant.quote(account, poolKey, true, amount, true, getMinSqrtPrice())
```
wojciech-cichocki, line 104: IMO, we can simplify the `invariant,quote` interface by removing `target_sqrt_price`. This parameter could be automatically set as either `min_sqrt_price` or `max_sqrt_price` depending on the swap direction (`xToY`).

sdk/src/index.ts
```diff
@@ -1,177 +1,164 @@
...
-  const blockNumber = await api.query.system.number()
-  for (let i = 0; i < 3; i++) {
-    const previousBlockNumber = (blockNumber as unknown as number) - 1 - i
-    const previousBlockHash = await api.query.system.blockHash(previousBlockNumber)
-    const apiAt = await api.at(previousBlockHash.toString())
-    const events = await apiAt.query.system.events()
-    console.log((events as any).length)
+  // throw an error if quote fails (it means that swap is not possible)
+  if (!quoteResult.ok) {
+    throw new Error(`quote returned an error: ${quoteResult.err?.toString()}`)
```
wojciech-cichocki, line 108: https://github.com/invariant-labs/protocol-a0/pull/187/files#r1465819320

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):
After merge this PR, align it with docs

wojciech-cichocki (approved):

PR #188, opened 2024-01-25
Refactor sdk

Sniezka1927:


sdk/math/wasm_helpers.rs
```diff
@@ -112,59 +109,43 @@ macro_rules! resolve {
...
+    upper_tick_fee_growth_outside_y: FeeGrowth,
+    pool_current_tick_index: i32,
+    pool_fee_growth_global_x: FeeGrowth,
+    pool_fee_growth_global_y: FeeGrowth,
+    position_fee_growth_inside_x: FeeGrowth,
+    position_fee_growth_inside_y: FeeGrowth,
+    position_liquidity: Liquidity,
+) -> TrackableResult<TokenAmounts> {
     let (fee_growth_inside_x, fee_growth_inside_y) = calculate_fee_growth_inside(
         lower_tick_index as i32,
```
wojciech-cichocki, line 128: Is casting to `as i32` still required here?

sdk/src/consts.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_REF_TIME = 1250000000000
```
wojciech-cichocki, line 1: Currently, there are two sources of constants: here and `.env`. We can eliminate all references to `.env`, such as `dotenv.config()`. Some constants are useful and should be transferred from `.env` to `consts.ts`. These constants include:

```bash
WAZERO_ADDRESS="5EFDb7mKbougLtr5dnwd5KDfZ3wK55JPGPLiryKq4uRMPR46"
ACCOUNT_URI="//Alice"
LOCAL="ws://localhost:9944"
CHAIN="alephzero-testnet"
```
Instead of loading from `.env` or using hardcoded values, we can import them from `consts.ts`

sdk/src/schema.ts
```diff
@@ -73,22 +79,21 @@ export type InvariantEventType =
 
 export interface TxResult {
   hash: string
-  events: InvariantEventType[]
 }
 
 export type CreatePositionTxResult = {
   hash: string
-  events: [CreatePositionEvent]
+  events: CreatePositionEvent[]
```
wojciech-cichocki, line 86: Are you sure you want to incorporate array here? The `create_position` entrypoint always emits one event in case of success.

Just reminder of the difference:
- `events: [CreatePositionEvent]` - This syntax is using tuple type. In TS, tuple is array with **fixed number** of elements whose types are known, but need not be the same. In this case, the tuple is defined to have exactly one element of type `CreatePositionEvent`. It's quite specific in its length and the type of its elements
- `events: CreatePositionEvent[]` - This syntax defines an array of `CreatePositionEvent` type. Unlike a tuple, this array can have **any number of elements** (including zero), but all elements must be of the `CreatePositionEvent` type. It's more flexible in terms of the number of elements it can contain.

wojciech-cichocki (changes requested):

sdk/src/utils.ts
```diff
@@ -212,15 +219,11 @@ export const calculateSqrtPriceAfterSlippage = (
...
-  return (
-    sqrt(
-      ((sqrtPrice * sqrtPrice) / getSqrtPriceDenominator()) *
-        multiplier *
-        getSqrtPriceDenominator() *
-        getPercentageDenominator()
-    ) / getPercentageDenominator()
-  )
+  const price = sqrtPriceToPrice(sqrtPrice)
+  const slippagePercentage =
```
wojciech-cichocki, line 223: Include `priceToSqrtPrice` function here:
```Typescript
export const calculateSqrtPriceAfterSlippage = (
  sqrtPrice: SqrtPrice,
  slippage: Percentage,
  up: boolean
): SqrtPrice => {
  const multiplier = getPercentageDenominator() + (up ? slippage : -slippage)
  const price = sqrtPriceToPrice(sqrtPrice)
  const priceWithSlippage = price * multiplier * getPercentageDenominator()
  const sqrtPriceWithSlippage = priceToSqrtPrice(priceWithSlippage) / getPercentageDenominator()
  
  return sqrtPriceWithSlippage
}
```

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):
Remove references to `process.env`.
So remove functions:
- getEnvTestAccount
- getEnvAccount

In the `initPolkadotApi` function, eliminate references to `process.env` and instead use `consts.ts`

sdk/math/wasm_helpers.rs
```diff
@@ -106,59 +106,43 @@ macro_rules! resolve {
...
+        .to_fee(position_liquidity))?;
     let tokens_owed_y = ok_or_mark_trace!(fee_growth_inside_y
-        .unchecked_sub(FeeGrowth::new(position_fee_growth_inside_y))
-        .to_fee(Liquidity::new(position_liquidity)))
-    .map_err(|e| e.to_string())?;
-
-    Ok(serde_wasm_bindgen::to_value(&TokenAmounts {
+        .unchecked_sub(position_fee_growth_inside_y)
+        .to_fee(position_liquidity))?;
+    Ok(TokenAmounts {
```
wojciech-cichocki, line 142: Is this struct still required? Can we replace it with tuple struct?

wojciech-cichocki (changes requested):

PR #193, opened 2024-01-27
refactor sdk

zielvna:


sdk/src/schema.ts
```diff
@@ -71,30 +70,16 @@ export type InvariantEventType =
...
   hash: string
-  events: [CrossTickEvent, SwapEvent] | [SwapEvent]
+  events: T
 }
 
-export type SwapRouteTxResult = {
-  hash: string
-  events: (CrossTickEvent | SwapEvent)[]
-}
+export type TxResult = TxResultInterface<InvariantEventType[]>
```
wojciech-cichocki, line 78: There should also be a result type for responses without events. While it's possible to use an empty array, this follows a soft-typed style.

I suggest using the `extends` interface to add an `events` field - this solves the problem, and we do not compromise with generics

`TxResult {hash}`
`EventTxResult extends TxResult {events}`

sdk/tests/math.test.ts
```diff
@@ -112,7 +112,8 @@ describe('check get liquidity by x', async () => {
         tokensOwedX: 0n,
         tokensOwedY: 0n
       }
-      await positionEquals(position, expectedPosition)
+      console.log(expectedPosition, position)
```
wojciech-cichocki, line 115: `console.log` leftover

wojciech-cichocki (changes requested):

sdk/src/utils.ts
```diff
@@ -96,7 +97,7 @@ export async function sendTx(
   data: any[],
   waitForFinalization: boolean = true,
   block: boolean = true
-): Promise<TxResult> {
+): Promise<EventTxResult<any>> {
```
wojciech-cichocki, line 96: `EventTxResult<any> | TxResult `?

sdk/src/utils.ts
```diff
@@ -24,27 +26,36 @@ import {
...
+    const chain = getSubstrateChain('alephzero-testnet')
+
     if (!chain) {
       throw new Error('chain not found')
     }
+
+    const { api } = await initApi(chain, { noInitWarn: true })
+    return api
+  } else if (network === Network.Mainnet) {
+    const chain = getSubstrateChain('alephzero-mainnet')
```
wojciech-cichocki, line 49: Use const for this and similar cases

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #198, opened 2024-02-06
move wasm and add it as workspace to sdk

zielvna:


wojciech-cichocki (changes requested):
Have you tried publishing this registry to test it?
I encountered an error during publishing: 'Hard link is not allowed'

wojciech-cichocki (changes requested):
What is the process of publishing now? I run `build.sh`, then `npm publish`, and still get error: "Hard link is not allowed"

wojciech-cichocki (approved):

PR #199, opened 2024-03-14
Updated PoolKeys struct

Sniezka1927:


src/contracts/collections/pool_keys.rs
```diff
@@ -1,41 +1,65 @@
...
-            .iter()
-            .position(|vec_pool_key| *vec_pool_key == pool_key)
-            .ok_or(InvariantError::PoolKeyNotFound)?;
+        let extracted_pool_key = self.get(pool_key);
+        if extracted_pool_key.is_none() {
+            return Err(InvariantError::PoolKeyNotFound);
+        }
 
-        self.pool_keys.remove(index);
+        let pool_key_index = extracted_pool_key.unwrap();
```
wojciech-cichocki, line 37: The syntax, when you first check `is_none()`/`is_some()` and then `unwrap` on the same object, can be replaced with a `match Some/None` pattern.

I'm not entirely certain if this case is correct for all edge cases, but here's a valid match Some/None schema:

```Rust
  match self.get(pool_key) {
      Some(pool_key_index) => {
          self.pool_keys.remove(&pool_key);
          self.pool_keys_by_index.remove(&pool_key_index);
          self.pool_keys_length -= 1;
          Ok(())
      },
      None => Err(InvariantError::PoolKeyNotFound),
  }
```

src/contracts/collections/pool_keys.rs
```diff
@@ -1,41 +1,65 @@
...
 
+        self.pool_keys.remove(pool_key);
+        self.pool_keys_by_index.remove(pool_key_index);
+        self.pool_keys_length -= 1;
         Ok(())
     }
 
-    pub fn contains(&self, fee_tier_key: PoolKey) -> bool {
-        self.pool_keys.contains(&fee_tier_key)
```
wojciech-cichocki, line 34: earlier implementation of this function looks strange. was there bug here previously?

src/contracts/collections/pool_keys.rs
```diff
@@ -1,41 +1,65 @@
...
+    pub fn contains(&self, pool_key: PoolKey) -> bool {
+        self.pool_keys.get(pool_key).is_some()
     }
 
-    pub fn get_all(&self) -> Vec<PoolKey> {
-        self.pool_keys.clone()
+    pub fn get_all(&self, size: u8) -> Result<Vec<PoolKey>, InvariantError> {
+        if size > 220 {
```
wojciech-cichocki, line 50: The best option is to save this value (`220`) as a constant with an explanation of the calculations, similar to the size limits we previously defined

src/e2e/create_pool.rs
```diff
@@ -306,4 +306,43 @@ pub mod e2e_tests {
...
+
+        let alice = ink_e2e::alice();
+
+        add_fee_tier!(client, InvariantRef, dex, fee_tier, alice).unwrap();
+
+        // size of PoolKeys = 32+32+8+2 = 74 bytes, max size of ink! storage cell is 16384 bytes
+        // that is 16384 - 32 (first 32B is the size of vec) = 16352 bytes left for memory
+        // 16352 / 74 = 220 elements. Adding 221st element will panic because of not enough memory in the storage cell
+
+        let amount_of_pools_to_create = 1000;
```
wojciech-cichocki, line 326: Have you tried the maximum value limited by length? `2^16 - 1 = 65535`

wojciech-cichocki (changes requested):

src/contracts/collections/pool_keys.rs
Sniezka1927: It was working correctly, the `self.pool_keys.contains(...)` was calling a vec
[method](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.contains)

wojciech-cichocki (approved):

PR #200, opened 2024-03-14
Add metadata

Sniezka1927:


sdk/tests/get-tickmap.test.ts
```diff
@@ -1,10 +1,10 @@
 import { Keyring } from '@polkadot/api'
 import { assert } from 'chai'
-import { Invariant } from '../src/invariant'
-import { Network } from '../src/network'
-import { PSP22 } from '../src/psp22'
-import { assertThrowsAsync } from '../src/testUtils'
-import { initPolkadotApi, integerSafeCast, newFeeTier, newPoolKey } from '../src/utils'
+import { Invariant } from '../src/invariant.js'
```
wojciech-cichocki, line 3: what is the reason for these changes in the file extensions?

src/e2e/mod.rs
```diff
@@ -1,24 +1,24 @@
-pub mod add_fee_tier;
-pub mod change_fee_receiver;
-pub mod change_protocol_fee;
-pub mod claim;
-pub mod constructor;
-pub mod create_pool;
-pub mod cross;
-pub mod cross_both_side;
+// pub mod add_fee_tier;
```
wojciech-cichocki, line 1: probably only for testing purposes

wojciech-cichocki (changes requested):

src/contracts/storage/tickmap.rs
```diff
@@ -249,11 +261,25 @@ impl Tickmap {
...
+
         self.bitmap.insert(
             (chunk, pool_key),
             &flip_bit_at_position(returned_chunk, bit),
         );
     }
+
+    pub fn store_initialized_chunk(&mut self, pool_key: PoolKey, chunk: u16) {
+        let mut vec = self.initialized_chunks.get(pool_key).unwrap_or_default();
+        vec.push(chunk);
```
wojciech-cichocki, line 276: Will these chunks not be saved multiple times here if I initialize multiple ticks in one chunk? (If this is true, then checking if the chunk is added may be too expensive)

wojciech-cichocki (changes requested):

wojciech-cichocki (changes requested):

PR #212, opened 2024-03-18
Cross testnet tests

Sniezka1927:


sdk/tests/crosses-testnet.test.ts
```diff
@@ -0,0 +1,78 @@
+import { Keyring } from '@polkadot/api'
+import { assert } from 'chai'
+import { isTokenX } from '../src/index.js'
+import { Invariant } from '../src/invariant.js'
+import { Network } from '../src/network.js'
+import { PSP22 } from '../src/psp22.js'
+import { ContractOptions } from '../src/schema.js'
+import { initPolkadotApi, integerSafeCast, newFeeTier, newPoolKey } from '../src/utils.js'
+
+describe('crosses', async () => {
```
wojciech-cichocki, line 10: can you exclude this test from the basic test e2e and let it be a separate command?

wojciech-cichocki: Also try to add tests with more detailed descriptions. For example: 'Validate limitation of number of crosses in atomic swap"

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #221, opened 2024-05-05
Wasm publish script

Sniezka1927:


sdk/publish-wasm.sh
```diff
@@ -0,0 +1,18 @@
...
+./package.sh
+
+jq '.name = "@invariant-labs/a0-sdk-wasm"' src/wasm/pkg/package.json > temp.json && mv temp.json src/wasm/pkg/package.json
+
+if [ -z "$1" ]; then
+    echo "Please provide the version to publish."
+    exit 1
+fi
+
+jq ".version = \"$1\"" src/wasm/pkg/package.json > temp.json && mv temp.json src/wasm/pkg/package.json
```
wojciech-cichocki, line 15: what about `"type": "module"`?

wojciech-cichocki (changes requested):

sdk/publish-wasm.sh
Sniezka1927: the type is set with the `./package.sh` script invoked before changing the name of wasm package

PR #225, opened 2024-05-10
add tx support

zielvna:


sdk/tests/tx.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  it('should send tx', async () => {
+    const mintTx = psp22.mintTx(500n)
+
+    const hash = await sendTx(mintTx)
+
+    expect(hash).to.not.be.undefined
+  })
+})
+
+async function sendTx(tx: SubmittableExtrinsic) {
```
wojciech-cichocki, line 30: Create a `sendTx` function that sends an already signed transaction and add this to SDK. The flow should be as follows:
1. Create a transaction using the SDK
2. Sign this transaction in the frontend
3. Send the transaction using the SDK

Here you doing both sign and send

sdk/src/invariant.ts
```diff
@@ -187,6 +188,17 @@ export class Invariant {
     )
   }
 
+  changeProtocolFeeTx(fee: Percentage) {
```
wojciech-cichocki, line 191: Add return type to function signature

sdk/src/wrapped-azero.ts
```diff
@@ -80,6 +80,24 @@ export class WrappedAZERO {
     )
   }
 
+  // approveTx(spender: string, value: bigint) {
```
wojciech-cichocki, line 83: leftover comment

sdk/src/invariant.ts
```diff
@@ -399,6 +506,12 @@ export class Invariant {
...
   }
 
+  claimFeeTx(index: bigint) {
+    return getTx(this.contract, this.gasLimit, this.storageDepositLimit, 0n, InvariantTx.ClaimFee, [
+      index
+    ])
+  }
+
   async claimFee(account: IKeyringPair, index: bigint, block: boolean = true): Promise<TxResult> {
     return sendTx(
```
wojciech-cichocki, line 516: Rename it to `sendAndSignTx` as this function doing both. Then you use name `sendTx` to function that only send signed transaction

sdk/src/utils.ts
```diff
@@ -85,6 +86,28 @@ export async function sendQuery(
   }
 }
 
+export function getTx(
```
wojciech-cichocki, line 89: IMO this function should be name `createTx`. The prefix `get` more implies that something already exists, you just want to gain access to it

wojciech-cichocki (changes requested):
Avoid defining the same thing multiple times. For example the same logic is used both case `getTx(...)`, and first part of `sendTx`


```Typescript
async function sendTx(...) {
  const tx = getTx(...)
  ...
}
```

sdk/src/utils.ts
```diff
@@ -109,6 +109,41 @@ export function createTx(
 }
 
 export async function sendTx(
+  tx: SubmittableExtrinsic,
+  waitForFinalization: boolean = true,
+  block: boolean = true
+): Promise<EventTxResult<any> | TxResult> {
+  return new Promise(async (resolve, reject) => {
```
wojciech-cichocki, line 116: same part logic as in `sendTx` and `signAndSendTx`. Use single point of definition

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #237, opened 2024-05-17
fix slippage tolerance and add slippage tolerance test

zielvna:


sdk/tests/position.test.ts
```diff
@@ -232,4 +234,38 @@ describe('position', async () => {
       assert.deepEqual(position.tokensOwedX, 0n)
     }
   })
+
+  it.only('slippage tolerance works', async () => {
```
wojciech-cichocki, line 238: remove `only`

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #239, opened 2024-05-17
export feeTiers array as const

p6te:
export feeTiers array as const


sdk/src/utils.ts
```diff
@@ -455,3 +456,7 @@ export const calculateLiquidityBreakpoints = (
     }
   })
 }
+
+export const generateFee = (tickCount: bigint): FeeTier => {
```
wojciech-cichocki, line 460: change naming 
```Typescript
export const calculateFeeTierWithLinearRatio = (tickSpacing: bigint): FeeTier => {
```

wojciech-cichocki (changes requested):
Also bump version in sdk, then run `build.sh` script

wojciech-cichocki (approved):

PR #246, opened 2024-05-24
Concentration utils

Sniezka1927:


sdk/src/utils.ts
```diff
@@ -598,8 +601,60 @@ export const calculateFeeTierWithLinearRatio = (tickCount: bigint): FeeTier => {
...
+) => {
+  const base = Math.pow(1.0001, -(tickSpacing / 4))
+  const logArg =
+    (1 - 1 / (concentration * CONCENTRATION_FACTOR)) /
+    Math.pow(1.0001, (-tickSpacing * minimumRange) / 4)
+
+  return Math.ceil(Math.log(logArg) / Math.log(base) / 2)
+}
+
+export const getConcentrationArray = (
```
wojciech-cichocki, line 628: Export already calculated arrays for all `FEE_TIERS`. it is always static, there is no point in counting it on the frontend

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #249, opened 2024-05-29
update get liquidity ticks related entrypoints

none00y:
# Changes
- update get_liquidity_ticks and get_liquidity_ticks_amount entrypoints
# Note
- the CI for the SDK appears to fail on the test `should get liquidity ticks with multiple queries` test, over multiple reruns  but it works fine locally. It might be that requests are sent faster in the github runner than on the local machine resulting in it's overloading


src/lib.rs
```diff
@@ -1113,21 +1083,67 @@ pub mod invariant {
...
+            if lower_tick < min_tick || upper_tick > max_tick {
+                return Err(InvariantError::InvalidTickIndex);
+            };
+
+            let (min_chunk_index, min_bit) = tick_to_position(lower_tick, tick_spacing);
+            let (max_chunk_index, max_bit) = tick_to_position(upper_tick, tick_spacing);
+
+            let active_bits_in_range = |chunk, min_bit, max_bit| {
+                let mut amount = 0;
+                for bit in min_bit..=max_bit {
```
wojciech-cichocki, line 1113: This does not look like an optimal function because it uses loops underneath. It is best to use the `count_ones` function.

The `count_ones()` leverages hardware instructions referred to as the "population count" instruction. POPCNT - Return the Count of Number of Bits Set to 1.

For example something similar to this:
```Rust
let active_bits_in_range = |chunk, min_bit, max_bit| -> u32 {
    ((chunk >> min_bit) & ((1 << (max_bit - min_bit + 1)) - 1)).count_ones()
};
```

src/lib.rs
```diff
@@ -1113,21 +1083,67 @@ pub mod invariant {
...
+            let max_chunk = self
+                .tickmap
+                .bitmap
+                .get((max_chunk_index, pool_key))
+                .unwrap_or(0);
+
+            let mut amount: u32 = 0;
 
-            let mut amount = 0;
+            amount += active_bits_in_range(min_chunk, min_bit, 63);
```
wojciech-cichocki, line 1137: Here you can even use `CHUNK_SIZE - 1` const instead magic number (just for code clarity)

src/contracts/entrypoints.rs
```diff
@@ -387,14 +387,25 @@ pub trait InvariantTrait {
...
     /// Retrieves the amount of liquidity ticks of a specified pool.
     ///
     /// # Parameters
     /// - `pool_key`: A unique key that identifies the specified pool.
+    /// - `lower_tick`: index to start counting from
+    /// - `upper_tick`: index to stop counting after(inclusive)
+    ///
     #[ink(message)]
-    fn get_liquidity_ticks_amount(&self, pool_key: PoolKey) -> u32;
+    fn get_liquidity_ticks_amount(
```
wojciech-cichocki, line 405: As I understand, there is always enough gas for this entrypoint, even in the worst-case scenario where the tick spacing is equal to 1 and the maximum search range is used? It's always worth including such information in the docs

wojciech-cichocki: I find this in SDK `// Query needs to be split in the case where tickSpacing = 1, otherwise a single query will fit within the gas limit`

sdk/src/invariant.ts
```diff
@@ -811,13 +817,24 @@ export class Invariant {
     )
   }
 
-  async getLiquidityTicksAmount(poolKey: PoolKey): Promise<bigint> {
-    return sendQuery(
+  // Query needs to be split in the case where tickSpacing = 1, otherwise a single query will fit within the gas limit
+  async getLiquidityTicksAmount(
```
wojciech-cichocki, line 821: Just a heads up: this function will only be used in an alternative scenario on the frontend (Of course, we'll keep this function)

sdk/src/invariant.ts
```diff
@@ -791,14 +791,20 @@ export class Invariant {
     return { bitmap: storedTickmap }
   }
 
-  async getLiquidityTicks(poolKey: PoolKey, offset: bigint): Promise<LiquidityTick[]> {
-    return sendQuery(
+  async getLiquidityTicks(poolKey: PoolKey, offset: bigint[]): Promise<LiquidityTick[]> {
```
wojciech-cichocki, line 794: rename argument `offset` to `ticks`

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #250, opened 2024-05-29
add withdraw all wazero entrypoint to contract and sdk

zielvna:


src/lib.rs
```diff
@@ -1129,6 +1132,28 @@ pub mod invariant {
...
+
+        #[ink(message)]
+        fn withdraw_all_wazero(&self, address: AccountId) -> Result<(), InvariantError> {
+            let caller = self.env().caller();
+            let contract = self.env().account_id();
+
+            let mut wazero_psp22: contract_ref!(PSP22) = address.into();
+            let mut wazero_wrapped_azero: contract_ref!(WrappedAZERO) = address.into();
+
+            let balance = wazero_psp22.balance_of(caller);
```
wojciech-cichocki, line 1144: Skip these invocations in case the balance equals 0

src/contracts/entrypoints.rs
```diff
@@ -397,4 +397,11 @@ pub trait InvariantTrait {
     /// - `pool_key`: A unique key that identifies the specified pool.
     #[ink(message)]
     fn get_liquidity_ticks_amount(&self, pool_key: PoolKey) -> u32;
+
+    /// Transfers, withdraws and transfers back all wAZERO tokens from the caller.
```
wojciech-cichocki, line 401: Expand on this description a bit, because it's not clear what the function is for. It's important to include that here as well:
- (1) implements unwrap_all_azero
- (2) invariant contract unwrap it on behalf of caller

sdk/src/invariant.ts
```diff
@@ -820,4 +820,22 @@ export class Invariant {
       [poolKey]
     )
   }
+
+  async withdrawAllWAZERO(
```
wojciech-cichocki, line 824: We also need the `withdrawAllWAZEROTx` version for frontend usage

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #251, opened 2024-05-31
add getAllLiquidityTicks

none00y:
# Changes
- add getAllLiquidityTicks to sdk


sdk/src/invariant.ts
```diff
@@ -807,6 +810,26 @@ export class Invariant {
...
+        }
+      }
+    }
+    const tickLimit = integerSafeCast(LIQUIDITY_TICKS_LIMIT)
+    const promises: Promise<LiquidityTick[]>[] = []
+    for (let i = 0; i < tickIndexes.length; i += tickLimit) {
+      promises.push(this.getLiquidityTicks(poolKey, tickIndexes.slice(i, i + tickLimit)))
+    }
+    
+    return Promise.all(promises).then(ticks => ticks.flat(1))
```
wojciech-cichocki, line 830: It would be better to use the newest syntax with `await`
Such like this:
```Typescript
  const tickResults = await Promise.all(promises);
  return tickResults.flat();
```

wojciech-cichocki (approved):

PR #267, opened 2024-06-14
fix swap simulation

zielvna:


sdk/tests/simulate-invariant-swap.test.ts
```diff
@@ -688,4 +689,44 @@ describe('simulateInvariantSwap', async () => {
     expect(simulation.stateOutdated).to.equal(false)
     expect(simulation.maxTicksCrossed).to.equal(true)
   })
+
+  it('max token amount', async () => {
```
wojciech-cichocki, line 693: You should also validate other case:
`xToY = {true, false}`  
`byAmountIn = {true, false}`

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #273, opened 2024-06-20
refactor get pools entrypoint

zielvna:


wojciech-cichocki (approved):
Looks good. On frontend level make you create only one fetch in case you did not exceed limit of poolkeys per single fetch. Optimistically there should be only one fetch, pessimistically a maximum of 2

PR #275, opened 2024-06-26
add get positions entrypoint

zielvna:


src/lib.rs
```diff
@@ -767,7 +767,35 @@ pub mod invariant {
...
+            offset: u32,
+        ) -> Result<(Vec<Position>, Vec<Pool>, Vec<Tick>, u32), InvariantError> {
+            let positions = self.positions.get_all(owner_id, size, offset);
+            let mut pools = vec![];
+            let mut ticks = vec![];
+            let positions_length = self.positions.get_length(owner_id);
+
+            for position in &positions {
+                pools.push(self.pools.get(position.pool_key)?);
+                ticks.push(
```
wojciech-cichocki, line 788: What about duplicated ticks? There are two ways to take advantage of them:
- Increase the limit of returning positions
- Use the type `Vec<(Pool, Tick, Tick)>`, which has a bit more consistency check here

src/contracts/collections/positions.rs
```diff
@@ -205,15 +213,17 @@ mod tests {
             ..Position::default()
         };
 
-        let result = positions.get_all(account_id);
+        let result = positions.get_all(account_id, 1, 0);
+        println!("{:?}", result);
```
wojciech-cichocki, line 217: leftover `println `

wojciech-cichocki (changes requested):

sdk/tests/get-positions.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    await invariant.createPosition(account, poolKey, -10n, 10n, 1000000000000n, pool.sqrtPrice, 0n)
+    await invariant.createPosition(account, poolKey, -20n, 20n, 1000000000000n, pool.sqrtPrice, 0n)
+  })
+
+  it('get positions', async () => {
+    const result = await invariant.getPositions(account.address, 2n, 0n)
+
+    assert.equal(result[0].length, 2)
+    assert.equal(result[1], 2n)
```
wojciech-cichocki, line 50: Check the details of the position, pool, and ticks to see if they match, at least for the first test. This will be useful as a regression test.

wojciech-cichocki (approved):

PR #279, opened 2024-07-04
implement seconds per liquidity type and update storage structs funct…

zielvna:
…ions

### DIFF
--- a/src/contracts/storage/position.rs
+++ b/src/contracts/storage/position.rs
@@ -6,6 +6,7 @@ use crate::{
         types::{
             fee_growth::{calculate_fee_growth_inside, FeeGrowth},
             liquidity::Liquidity,
+            seconds_per_liquidity::{calculate_seconds_per_liquidity_inside, SecondsPerLiquidity},
             sqrt_price::SqrtPrice,
             token_amount::TokenAmount,
         },
@@ -26,6 +27,7 @@ pub struct Position {
     pub last_block_number: u64,
     pub tokens_owed_x: TokenAmount,
     pub tokens_owed_y: TokenAmount,
+    pub seconds_per_liquidity_inside: SecondsPerLiquidity,
 }
 
 impl Position {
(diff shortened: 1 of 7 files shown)


src/contracts/storage/position.rs
```diff
@@ -193,16 +198,18 @@ impl Position {
...
 
+        let current_timestamp_in_seconds = current_timestamp_in_milliseconds / 1000;
         let (required_x, required_y) = unwrap!(position.modify(
             pool,
             upper_tick,
             lower_tick,
             liquidity_delta,
             true,
-            current_timestamp,
+            current_timestamp_in_seconds,
```
wojciech-cichocki, line 212: There is no place where we need a timestamp in milliseconds, so it’s worth modifying the function that retrieves the timestamp to return it in seconds to avoid conversions and potential errors.
`get_timestamp`

src/lib.rs
```diff
@@ -391,6 +391,10 @@ pub mod invariant {
             self.env().block_timestamp()
         }
 
+        fn get_timestamp_in_seconds(&self) -> u64 {
```
wojciech-cichocki, line 394: There should be only one use for getting the timestamp

wojciech-cichocki (changes requested):
The function for calculating the maximum number of crosses should be updated to check crosses in the pessimistic case

sdk/tests/crosses-testnet.test.ts
```diff
@@ -43,20 +43,20 @@ describe('testnet-crosses-limitations', async () => {
...
+        -i * 127n,
         liquidityDelta,
         spotSqrtPrice,
         slippageTolerance
       )
     }
 
     const swapper = keyring.addFromUri('//Bob')
-    const swapAmount = 909000n
+    const swapAmount = 418000n
```
wojciech-cichocki, line 59: It's difficult to test with a constant amount, because each swap shifts the price, and the amount needed to cross changes. Ideally, this test would retrieve the state of ticks and calculate the amount needed to cross based on that - it's quite simple since the same logic is already present in the swap

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #280, opened 2024-07-04
refactor get all pool keys and get all positions

zielvna:


sdk/src/invariant.ts
```diff
@@ -528,6 +530,37 @@ export class Invariant {
       refTime: this.gasLimit.refTime.toNumber(),
       proofSize: this.gasLimit.proofSize.toNumber()
     }
+  ): Promise<[Position, Pool, Tick, Tick][]> {
+    const [positions, poolKeysCount] = await this.getPositions(
```
wojciech-cichocki, line 534: IMO, there should be an option to skip the first fetch and consequently pass positionCount. This way, we can decrease the time and avoid the sequence of two fetches. 

At the moment of typing something in the search, you know how many positions there are and you probably also have one of the pages (at least).

wojciech-cichocki (changes requested):

sdk/src/invariant.ts
```diff
@@ -522,6 +527,72 @@ export class Invariant {
...
+    let actualPositionsCount = positionsCount
+    if (!positionsCount) {
+      const [positionEntries, positionsCount] = await this.getPositions(
+        owner,
+        POSITIONS_ENTRIES_LIMIT,
+        BigInt(firstPageIndex - 1) * POSITIONS_ENTRIES_LIMIT,
+        options
+      )
+
+      pages.push({ index: 1, entries: positionEntries })
```
wojciech-cichocki, line 557: Why do you count pages from index 1?

sdk/src/invariant.ts
```diff
@@ -522,6 +527,72 @@ export class Invariant {
...
+    positionsCount?: bigint,
+    skipPages?: number[],
+    options: ContractOptions = {
+      storageDepositLimit: this.storageDepositLimit,
+      refTime: this.gasLimit.refTime.toNumber(),
+      proofSize: this.gasLimit.proofSize.toNumber()
+    }
+  ): Promise<Page[]> {
+    let firstPageIndex = 0
+    for (let i = 1; i < Number.MAX_SAFE_INTEGER; i++) {
```
wojciech-cichocki, line 540: More in this direction, it scales significantly better:
```Typescript
firstPageIndex = skipPages?.find(i => !skipPages.includes(i)) || 1;
```

btw try to ask chatGPT more often about possible refactorings

wojciech-cichocki (changes requested):

wojciech-cichocki (approved):

PR #281, opened 2024-07-08
U256T PoC

Sniezka1927:


src/lib.rs
```diff
@@ -60,11 +60,12 @@ pub mod invariant {
 
     impl Invariant {
         #[ink(constructor)]
-        pub fn new(protocol_fee: Percentage) -> Self {
+        pub fn new(protocol_fee: Percentage, poc_field: PocType) -> Self {
```
wojciech-cichocki, line 63: also, check how to pass this argument from the TS level and whether it really works

wojciech-cichocki (changes requested):

PR #283, opened 2024-07-10
add set code message to contract and sdk

zielvna:


wojciech-cichocki (commented):
It would be good to verify using ink e2e tests an update attempt from a non-admin account and from an admin account

wojciech-cichocki (approved):

PR #287, opened 2024-07-26
Remove deprecated entrypoints

Sniezka1927:


src/contracts/storage/tickmap.rs
```diff
@@ -25,6 +25,7 @@ impl Default for Tickmap {
     }
 }
 
+#[allow(dead_code)]
 pub fn get_max_chunk(tick_spacing: u16) -> u16 {
```
zielvna, line 29: can it be removed?

src/contracts/collections/pool_keys.rs
```diff
@@ -34,38 +30,25 @@ impl PoolKeys {
...
         let offset_with_size = offset.checked_add(size).unwrap();
 
-        let max = if offset_with_size > self.pool_keys_length {
+        let upper_bound = if offset_with_size > self.pool_keys_length {
             self.pool_keys_length
         } else {
             offset_with_size
         };
 
+        let max = if upper_bound.checked_sub(offset).unwrap() > MAX_POOL_KEYS_RETURNED {
```
zielvna, line 46: same thing with `get_all` function in positions collection would be helpful

src/contracts/storage/pool_key.rs
```diff
@@ -5,6 +5,8 @@ use crate::contracts::FeeTier;
 use crate::contracts::InvariantError;
 use crate::math::percentage::Percentage;
 
+pub const MAX_POOL_KEYS_RETURNED: u16 = 220;
```
zielvna, line 8: maybe something more dynamic would be better: `MAX_RESULT / POOL_KEY_SIZE` (like we have it in wasm)

src/contracts/collections/pool_keys.rs
wojciech-cichocki: Yes, we should add this to the `get_all` function in the `positions` collection as well

src/contracts/storage/tickmap.rs
wojciech-cichocki: We should keep this function as it's used in e2e tests. `#[allow(dead_code)]` is needed to remove the warning during the contract build

src/contracts/storage/pool_key.rs
wojciech-cichocki: definitely

zielvna (approved):

wojciech-cichocki (approved):

PR #290, opened 2024-07-31
Position with associates

DevRozaDev:
\+ remove `PositionTick` struct & `get_position_ticks` entrypoint
\+ add `get-position-with-associates` entrypoint & to sdk


src/contracts/entrypoints.rs
```diff
@@ -382,20 +382,27 @@ pub trait InvariantTrait {
...
     #[ink(message)]
-    fn get_position_ticks(&self, owner: AccountId, offset: u32) -> Vec<PositionTick>;
+    fn get_user_position_amount(&self, owner: AccountId) -> u32;
 
-    /// Retrieves the amount of positions held by the user.
+    /// Retrieves information about a single position, the associated pool, lower and upper tick in this order.
     ///
     /// # Parameters
-    /// - `owner`: An `AccountId` identifying the user who owns the position.
+    /// - `owner_id`: An `AccountId` identifying the user who owns the position.
```
zielvna, line 395: probably it should be just `owner` here

sdk/tests/get-position-with-associates.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      const [position, pool, lowerTick, upperTick] = await invariant.getPositionWithAssociates(account.address, 0n)
+    
+      assert.deepEqual(position, positionRegular)
+      assert.deepEqual(pool, poolRegular)
+      assert.deepEqual(lowerTick, lowerTickRegular)
+      assert.deepEqual(upperTick, upperTickRegular)
+    })
+
+    it('position does not exist', async () => {
+        assertThrowsAsync(invariant.getPositionWithAssociates(account.address, 0n), InvariantError.PositionNotFound)
```
zielvna, line 88: you need to await `assertThrowsAsync`, otherwise the test will not wait for result of the call and always pass

zielvna (changes requested):

Sniezka1927 (approved):

wojciech-cichocki (approved):

PR #302, opened 2024-09-27
Update get tickmap entrypoint

none00y:
# Changes
- add chunk lookups to get tickmap entrypoint
- update getFullTickmap to reduce query count


wojciech-cichocki (changes requested):
Please also validate the case where you return the maximum size of initialized tick indexes that are evenly distributed in tickmap

DevRozaDev (changes requested):
Please remove the `console.log`s from tests

wojciech-cichocki (approved):

PR #305, opened 2024-10-08
Preaudit changes

DevRozaDev:
\+ rename `InvariantTrait` to `InvariantEntrypoints`
\+ split  `route` into `route` and `route_mut` to make `quote_route` take an immutable reference
\+ `get_position` and `get_positions` take an immutable reference
\+ require passing the correct address to `withdraw_all_wazero` (`5CtuFVgEUz13SFPVY6s2cZrnLDEkxQXc19aXrNARwEBeCXgg`) [source](https://github.com/Cardinal-Cryptography/wAZERO)
\+ add `dev` feature that removes the requirement of passing mainnet's  `withdraw_all_wazero` address


Cargo.toml
```diff
@@ -19,6 +19,7 @@ scale-info = { version = "2.11.3", default-features = false }
 [dev-dependencies]
 ink_e2e = "5.0.0"
 test_helpers = { path = "src/test_helpers", default-features = false }
+scale-info = "2.11.3"
```
wojciech-cichocki, line 22: what's the reason for adding this `dev-dependencies`?

wojciech-cichocki (changes requested):

Cargo.toml
DevRozaDev: rust-analyzer gets fussy without it for me locally, my bad will remove

wojciech-cichocki (approved):

